PR #5, opened 2025-04-28
fix: fix group drag and drop loosing data during the drag drop process

mobaetzel:
If moving groups from one segment into another in another group, data was dropped from the destination group during this process. This was due to a lack of typing and missing data while passing the updated parent elements around.


jp-performance:
Task linked: [CU-86c38yza4 Gruppe kann nicht zwischen Abschnitten verschoben werden](https://app.clickup.com/t/86c38yza4)

PR #22, opened 2025-07-10
feat: SBOM generation and license compliance check

danielkratz:
Adds generation of SBOM (Software Bill of Materials) in CycloneDX format to build workflow and implements a license compliance check based on the generated SBOM, which generates a comment on the corresponding pull request. 

---

This PR includes a high number of commits and force-pushes due to the exploratory nature of the work.

The goal was to build a reliable and maintainable SBOM generation workflow and license compliance check based on the generated CycloneDX SBOM.

Since GitHub Actions does not offer a proper local test environment, multiple push cycles were required to verify the behavior of the automation.

Now that the implementation is stable and verified, the history may be a bit noisy, but the end result is clean and reproducible.


.github/workflows/build-and-push.yml
```diff
@@ -92,6 +92,148 @@ jobs:
...
+          node-version: '20'
+
+      - name: Prepare sbom output directory
+        run: mkdir -p app/public/sbom/partials
+
+      - name: Update package.json versions
+        run: |
+          echo "Injecting version: ${{ env.BUILD_VERSION }}"
+          jq --arg v "${{ env.BUILD_VERSION }}" '.version = $v' app/package.json > app/tmp.json && mv app/tmp.json app/package.json
+          jq --arg v "${{ env.BUILD_VERSION }}" '.version = $v' mails/package.json > mails/tmp.json && mv mails/tmp.json mails/package.json
```
mobaetzel, line 126: This process updates the version in `package.json` but does not update `pom.xml`. Since the versions are not synchronized, this step may not be necessary. If it's not required for our workflow, we can consider skipping it to avoid version inconsistencies.

mobaetzel (approved):

.github/workflows/build-and-push.yml
danielkratz: Thanks for bringing this up! We need to set the version for the `gover-frontend`. `gover-backend` and `gover-mail-templates` packages which will show up in the SBOM. For the backend we don't set the version in the `pom.xml`, but we set the project version temporarily via `mvn versions:set -DnewVersion=${{ env.BUILD_VERSION }}` so it gets generated correctly. In conclusion it is in fact required for our new workflow with the SBOM generation.

mobaetzel (approved):

PR #36, opened 2025-07-23
feat: new simplified versioning for forms and presets

danielkratz:
This pull request introduces a simplified versioning mechanism for both forms and presets. Instead of relying on manually entered semantic versions (1.0.0, 1.2.3, etc.), new versions are now automatically assigned using incremental whole numbers (1, 2, 3, ...). This applies when creating a new form version or a new preset version.

Legacy semantic versions remain fully supported and are preserved as-is.


mobaetzel (commented):
As discussed in our daily, we should rework the forms with explicite parent objects which the different versions relate to.

PR #38, opened 2025-07-24
patch: 4.5.2

mobaetzel:
# Description
- fix a bug in the preset derivation process


jp-performance (approved):

jp-performance (changes requested):
Changelog entry still refers to 4.5.1 instead of 4.5.2 and release date has to be set.

jp-performance (approved):

PR #42, opened 2025-07-27
build: integrate gover-next package in build pipeline

danielkratz:
This PR changes our build pipeline so that pre-Release builds are now published to a separate 'gover-next' package.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -20,8 +20,8 @@ on:
 env:
   REGISTRY: ghcr.io
   PR_NUMBER: ${{ github.event.pull_request.number }}
-  IS_ALPHA_RELEASE: ${{ github.event_name == 'pull_request' && github.head_ref != 'main' }}
-  IS_BETA_RELEASE: ${{ github.event_name == 'pull_request' && github.head_ref == 'main' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -15,10 +15,10 @@ on:
...
       - "patch/*"
       - "main"
+      - "next"
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: ${{ github.repository }}
   PR_NUMBER: ${{ github.event.pull_request.number }}
   IS_ALPHA_RELEASE: ${{ github.event_name == 'pull_request' && github.head_ref != 'main' }}
   IS_BETA_RELEASE: ${{ github.event_name == 'pull_request' && github.head_ref == 'main' }}
```
mobaetzel: I think, this mechanism does not work with the next branch. milestone/patch releases to the next branch will always be alpha releases. Is this behaviour correct?

danielkratz: You are absolutely right. That's an oversight from me. I fixed it so PRs to both main and next are correctly labeled as Beta. I also updated the determination of `IS_ALPHA_RELEASE` and `IS_BETA_RELEASE` to use `base_ref` instead of `head_ref` since we decide based on the target branch. It also seems like we don't use these environment variables anymore, since the version label is determined in the "Extract clean version tag".  This was already true for the prior version.

mobaetzel (approved):

PR #59, opened 2026-01-29
fix: destination url pattern

mobaetzel:


app/src/modules/destination/pages/details/destination-details-page-index.tsx
danielkratz, line 1: The URL validation was deactivated since it does not allow query params on the URL. We should keep the validation and modify the RegEx to allow params on the URL. 
Maybe something like this which allows anything after the host and optional port? 
/^https?:\/\/[\w.-]+(:\d+)?([^\s]*)$/

src/main/java/de/aivot/GoverBackend/services/DestinationSubmitService.java
```diff
@@ -145,13 +152,44 @@ private DestinationResponse sendHttp(Destination destination, Form form, Submiss
         } catch (MalformedURLException e) {
             return new DestinationResponse(false, "Die URL des Ziels ist ungültig", null, null);
         }
-        HttpURLConnection con;
+        HttpsURLConnection con;
         try {
-            con = (HttpURLConnection) url.openConnection();
+            con = (HttpsURLConnection) url.openConnection();
```
danielkratz, line 185: If we allow http URLs, which the (old) RegEx does (see other comment), this might lead to an error.

danielkratz (commented):
Please check the two comments I left.

app/src/modules/destination/pages/details/destination-details-page-index.tsx
mobaetzel: Thanks for the regex, it is implemented and the client side validation was reactivated

danielkratz (approved):

PR #60, opened 2026-02-12
feat: OZG-Cloud Destination

mobaetzel:
# Description
There is a need to send data to the OZG-Cloud. This should be done via a new destination type. This feature implements this new type in the front- and backend.

The corresponding Story is:
CU-86c869n6m

This PR includes the following tasks:
- CU-86c86f836
- CU-86c86f891
- CU-86c86f81q


app/src/modules/destination/pages/details/destination-details-page-index.tsx
```diff
@@ -452,6 +464,53 @@ export function DestinationDetailsPageIndex() {
...
+
+                    <Typography
+                        sx={{
+                            mb: 1,
+                        }}
+                    >
+                        Konfigurieren Sie die OZG-Cloud-Schnittstelle, an die Formulare gesendet werden sollen.
+                    </Typography>
+
+                    <OzgCloudInfo />
```
danielkratz: I would maybe switch the order here and show the endpoint setting before the notice on the technical field. That's because the endpoint is the setting and therefore a primary action. The note ist informational and is shown again in the form setting when configuring the destination.

danielkratz (approved):

danielkratz (approved):

PR #64, opened 2026-02-24
feature: new permission editor

danielkratz:


app/src/components/permission-editor/permission-editor.tsx
```diff
@@ -0,0 +1,792 @@
...
+                    ...PermissionGroups.map((group) => ({
+                        contextLabel: group.label,
+                        permissions: group.permissions.map((per) => ({
+                            label: PermissionLabelsDe[per],
+                            permission: per,
+                            description: '',
+                        })),
+                    })),
+                    ...scopedApiPermissions,
+                ] as PermissionDetails[];
```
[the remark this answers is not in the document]
danielkratz: @mobaetzel That's a known issue we have to talk about. The permission groups are not scoped (yet), so I only filter the permissions from the PermissionApiService at the moment.

app/src/components/permission-editor/permission-editor-props.tsx
```diff
@@ -0,0 +1,25 @@
+export interface PermissionDetails {
+    contextLabel: string;
+    permissions: Array<{
+        permission: string;
+        label: string;
+        description: string;
+    }>;
```
mobaetzel: There is already an interface for this dataset: PermissionEntry in the file app/src/modules/permissions/models/permission-provider.ts

app/src/components/permission-editor/permission-editor-props.tsx
```diff
@@ -0,0 +1,25 @@
...
+    /** Controlled change callback. */
+    onChange: (next: string[]) => void;
+    /** When false, disables selection and bulk actions. */
+    isEditable?: boolean;
+    /** When true, disables inputs and copy action. */
+    isBusy?: boolean;
+    /** Optional label above editor. */
+    title?: string;
+    /** Scope of the permissions. */
+    scope?: 'System' | 'Domain' | Array<'System' | 'Domain'>;
```
mobaetzel: There is already a type for this kind of data: PermissionScope in the file app/src/modules/permissions/enums/permission-scope.ts

app/src/components/permission-editor/permission-editor.tsx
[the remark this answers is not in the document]
mobaetzel: We will migrate all permissions onto the backend side permission providers. these are all scoped. the client side dataset will be removed in the future

app/src/components/permission-editor/permission-editor.tsx
```diff
@@ -0,0 +1,792 @@
...
+
+export function PermissionEditor(props: PermissionEditorProps): React.ReactElement {
+    const {
+        originalPermissions = [],
+        value,
+        onChange,
+        isEditable = true,
+        isBusy = false,
+        title = 'Berechtigungen',
+        scope = ['System', 'Domain'],
```
mobaetzel: There is a type for this: PermissionScope

app/src/modules/permissions/components/permission-editor.tsx
```diff
@@ -0,0 +1,792 @@
...
+    }, [originalPermissions, selectedPermissions]);
+
+    const filteredPermissionGroups = useMemo(() => {
+        const q = normalizeSearch(permissionQuery);
+        if (!q) {
+            return permissions.map((g) => ({
+                ...g,
+                permissions: [...g.permissions].sort((a, b) => a.label.localeCompare(b.label)),
+            }));
+        }
```
mobaetzel, line 206: We have the package fuse.js installed for a fuzzy search. Maybe this would be useful here

app/src/components/permission-editor/permission-editor.tsx
```diff
@@ -0,0 +1,792 @@
...
+
+    const expandAll = (next: boolean, scope: 'all' | 'filtered' = 'all'): void => {
+        const base = scope === 'filtered' ? filteredPermissionGroups : permissions;
+        const nextExpanded: Record<string, boolean> = {};
+        for (const g of base) nextExpanded[groupKey(g.contextLabel)] = next;
+
+        setExpandedGroups((prev) => ({...prev, ...nextExpanded}));
+    };
+
+    const expandSmart = (): void => {
```
mobaetzel: What does "smart" mean here? Maybe the name is not explicite enough.

app/src/modules/permissions/components/permission-editor.tsx
```diff
@@ -0,0 +1,792 @@
...
+            <Stack spacing={2}>
+                {filteredPermissionGroups.map((group) => {
+                    const total = group.permissions.length;
+                    const selectedInGroup = group.permissions.filter((p) => selectedPermissions.includes(p.permission)).length;
+                    const allSelected = total > 0 && selectedInGroup === total;
+                    const someSelected = selectedInGroup > 0 && selectedInGroup < total;
+
+                    const isExpanded = expandedGroups[groupKey(group.contextLabel)] ?? false;
+
+                    return (
```
mobaetzel, line 520: This large function might be a good candidate to be refactored into its own component.

app/src/modules/users/pages/user/details/user-details-page-index.tsx
mobaetzel: These roles might need to be fetched and the user entity field systemRoleId should be used to identify the system role

app/src/components/permission-editor/permission-editor-props.tsx
mobaetzel, line 1: it might be a better approach to place this permission editor in the corresponding permissions module to follow our target project structure where logic regarding a single specific topic lives in the concerned module. Additionally, we have started to internalize props interfaces which are not shared among multiple components. maybe this interface can be included into the main component file.

app/src/components/permission-editor/permission-editor.tsx
```diff
@@ -0,0 +1,655 @@
...
+            document.body.appendChild(textarea);
+            textarea.focus();
+            textarea.select();
+            document.execCommand('copy');
+            document.body.removeChild(textarea);
+            dispatch(showSuccessSnackbar('Permission-Key kopiert.'));
+        } catch {
+            dispatch(showErrorSnackbar('Kopieren nicht möglich.'));
+        }
+    };
```
mobaetzel: Maybe we can move this function into the utility package and reuse it wherever possible because we do this quiet often and a unified approach would be a good idea.

app/src/components/permission-editor/permission-editor.tsx
```diff
@@ -0,0 +1,655 @@
...
+                          <IconButton
+                              aria-label="Alle einklappen"
+                              size="small"
+                              onClick={() => expandAll(false, 'all')}
+                              disabled={isBusy || permissions.length === 0}
+                          >
+                            <UnfoldLessIcon fontSize="small" />
+                          </IconButton>
+                        </span>
+                    </Tooltip>
```
mobaetzel: It might be possible to reuse the actions component here `app/src/components/actions/actions.tsx`

app/src/components/permission-editor/permission-editor.tsx
```diff
@@ -0,0 +1,655 @@
...
+                            size="small"
+                            onClick={() => setPermissionQuery('')}
+                            aria-label="Suche löschen"
+                        >
+                            <ClearIcon fontSize="small" />
+                        </IconButton>
+                    ) : undefined,
+                }}
+                sx={{mt: 1.5, mb: 2}}
+            />
```
mobaetzel: We have two search field implementations, `app/src/components/search-input` and `app/src/components/search-input-2`. Maybe we can reuse. They are bound to be unified in the future and version 2 is slightly more standarized so I would suggest using this.

app/src/components/permission-editor/permission-group-accordion.tsx
```diff
@@ -0,0 +1,237 @@
...
+                                    py: 1,
+                                    borderRadius: 2,
+                                    display: 'flex',
+                                    alignItems: 'flex-start',
+                                    gap: 1,
+                                    '&:hover': {backgroundColor: 'action.hover'},
+                                }}
+                            >
+                                <FormControlLabel
+                                    sx={{m: 0, flex: 1, alignItems: 'flex-start'}}
```
mobaetzel: For easier git operations, js objects and arrays should be multiline as a rule of thumb. this drastically eases future extensions of said objects and arrays and merging them via git because git is line based. Maybe we can include this in the default IDE settings and eslint.

danielkratz: I added a ESLint warning for now, so we can align the project as a whole in a separate PR in the future and enforce the corresponding directives.

app/src/components/permission-editor/permission-editor.tsx
danielkratz: I merged the two components together with new options and logic for the new search field and made sure to use the component where it makes sense.

app/src/components/element-editor-code-tab/override-code-tab.tsx
```diff
@@ -167,8 +168,13 @@ export function OverrideCodeTab(props: OverrideCodeTabProps) {
 
                         dispatch(showSuccessSnackbar('Element-ID eingefügt'));
                     } else {
-                        navigator.clipboard.writeText(element.id);
-                        dispatch(showSuccessSnackbar('Element-ID kopiert'));
+                        copyToClipboardText(element.id).then((success) => {
+                            if (success) {
+                                dispatch(showSuccessSnackbar('Element-ID kopiert'));
+                            } else {
+                                dispatch(showSuccessSnackbar('Element-ID konnte nicht kopiert werden'));
```
mobaetzel: I think this should be an `showErrorSnackbar`

app/src/components/search-input/search-input.tsx
```diff
@@ -1,75 +1,140 @@
-import {styled} from '@mui/material/styles';
-import {InputBase} from '@mui/material';
+import {Box, IconButton, InputAdornment, SxProps, TextField} from '@mui/material';
 import React from 'react';
-import {SearchInputProps} from './search-input-props';
-import SearchOutlinedIcon from '@mui/icons-material/SearchOutlined';
+import ClearIcon from '@mui/icons-material/Clear';
+import SearchIcon from '@mui/icons-material/Search';
```
mobaetzel: Here are MUI Icons instead of the new Material Symbols.

app/src/components/copy-to-clipboard-button/copy-to-clipboard-button.tsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import {IconButton, SxProps, Theme, Tooltip} from '@mui/material';
+import ContentCopy from '@mui/icons-material/ContentCopyOutlined';
```
mobaetzel: You are using a MUI Icon instead of Material Symbols.

PR #68, opened 2026-03-05
feat: improve details for permissions

mobaetzel:
# Description
With the existing permission views it is hard to tell if a permission is direct because the user is a member of the department/team or indirect because the user is a deputy for a direct member. Sometimes both statements are true, a user is a direct member and the user is a deputy and receives indirect permissions. We need to be able to distinguish between these types of permissions because we want to only users for the process task assignment with direct permissions to be selectable but all users with direct or indirect permissions to be able to work on the tasks if they have the corresponding permissions.

This PR should implement distinguishable direct and indirect permissions as well as the information if the user has this permissions because the user is a direct or indirect member.


mobaetzel:
@danielkratz maybe we can review this together with your upcoming feature for the process task user select.

danielkratz (approved):

PR #71, opened 2026-03-05
feat: prototype consolidation

mobaetzel:
Collect all features for the upcoming prototype.


mobaetzel (commented):
test comment

mobaetzel (commented):
# Code Review: feat/prototype-consolidation → milestone/5.0.0

> **Scope:** 868 files changed · 61,851 insertions · 13,091 deletions · 483 commits
> **Reviewer:** Claude Code (automated deep-review across all layers)

---

## Overview

This PR consolidates all prototype features into the 5.0.0 milestone. It represents a **major architectural maturation** of the platform — not just feature additions, but a systematic redesign of:

- The process node execution engine (new plugin architecture + semantic versioning)
- The element data model (`ElementData` → `AuthoredElementValues` + `DerivedRuntimeElementData`)
- The permissions system (centralized → decentralized per-module `PermissionProvider`)
- The audit logging system (ad-hoc logging → fluent builder with DB persistence + diffs)
- The deployment/installation story (new `install.sh`, healthcheck, plugin loading)

Overall quality is **high**. The changes are well-structured, well-tested, and show clear architectural intent. Findings below are grouped by severity.

---

## 🔴 Critical / Blockers

### 1. Export Signature Field is a TODO — Do Not Ship to Production

**File:** `src/main/java/de/aivot/GoverBackend/process/services/ProcessNodeExportService.java`

The export payload includes a `signature` field currently set to `"TODO"`. If the import endpoint trusts this signature to validate the integrity of imported node definitions, this is a security gap. Unsigned exports can be tampered with before re-import.

**Action required:** Either implement cryptographic signing (HMAC or asymmetric) before enabling the import endpoint in production, or clearly gate the import feature behind a flag and document that signature validation is not yet implemented.

**Fixed by @mobaetzel on 24.03.2026 06:34**

---

### 2. `install.sh` Generates a `docker-compose.yml` with Keycloak Using Wrong Hostname Config

**File:** `install.sh` (lines ~190–210)

The generated `docker-compose.yml` inside `install.sh` configures Keycloak with:
```yaml
KC_HOSTNAME: ${HOSTNAME}
KC_HOSTNAME_PATH: /idp
```

But the main `docker-compose.yml` was updated (in this very PR) to use the Keycloak v2 format:
```yaml
KC_HOSTNAME: "https://{{ HOSTNAME }}/idp"
# KC_HOSTNAME_PATH removed
```

The `install.sh`-generated file is **out of sync** with the main `docker-compose.yml` and with the fix from PR #77. This will cause Keycloak startup failures for new installations using `install.sh`.

**Action required:** Align the Keycloak config in `install.sh` with the v2 hostname format.

**Fixed by @mobaetzel on 24.03.2026 06:37**

---

## 🟠 High Severity / Should Fix Before Merge

### 3. `ProcessNodeEntity` — Validation Message Not Updated After Size Increase

**File:** `src/main/java/de/aivot/GoverBackend/process/entities/ProcessNodeEntity.java`

The `processNodeDefinitionKey` field was changed from `max = 32` to `max = 128`, but the error message was not updated and still says `"32 Zeichen"` (or similar). Users will receive a misleading validation error.

**Fixed by @mobaetzel on 24.03.2026 06:40**

---

### 4. Audit Diff Calculation May Silently Miss Nested Changes

**File:** `src/main/java/de/aivot/GoverBackend/audit/models/AuditLogPayload.java`

The `createDiff()` method compares old/new `Map` states. If any entity serializes a nested object to a non-`Map` type (e.g., a `List` containing `Map`s, or a custom object), the recursive comparison will fall through to object equality rather than field-level diffing. This means some configuration changes in complex node entities may appear as `{old: {...}, new: {...}}` opaque blobs in the audit log rather than per-field diffs.

**Recommendation:** Add test cases for nested list-of-maps and mixed-type structures in the diff utility.

**Fixed by @mobaetzel on 24.03.2026 07:02**

---

### 5. `ProcessInstanceRetentionCleanupService` — No Batch Limit on Deletion

**File:** `src/main/java/de/aivot/GoverBackend/processinstance/services/ProcessInstanceRetentionCleanupService.java`

The cleanup service loops through **all** completed instances past their `keepUntil` date and deletes them one by one. In a production system that has accumulated thousands of expired instances, the first run after deployment will attempt to delete them all in a single nightly job, which may:
- Lock tables for an extended period
- Cause OOM or timeouts if instances have large attached data
- Run past the next scheduled window

**Recommendation:** Add a batch size limit (e.g., delete max 500 per run) and log how many were cleaned up vs. how many remain.

**Fixed by @mobaetzel on 24.03.2026 07:12**

---

### 6. Upload Size Limits Increased 5×/2× Without Documented Rationale

**File:** `src/main/resources/application.yml`

```yaml
# Before
max-file-size: 10MB
max-request-size: 100MB

# After
max-file-size: 50MB
max-request-size: 200MB
```

A 5× increase in per-file limit is significant. This should be:
- Documented in the changelog
- Validated against the antivirus (ClamAV) service's default scan limits
- Validated against Kubernetes/container memory limits if deployed in such environments

**Checked by @mobaetzel on 24.03.2026 07:21:** The max file size in the default clamav debian container seems to be 100MB. The excerpt from the container clamd.conf at `/etc/clamav/clamd.conf`:

```
# This option sets the maximum amount of data to be scanned for each input
# file. Archives and other containers are recursively extracted and scanned
# up to this value.
# Value of 0 disables the limit
# Note: disabling this limit or setting it too high may result in severe damage
# to the system.
# Default: 400M
#MaxScanSize 1000M

# Files larger than this limit won't be scanned. Affects the input file itself
# as well as files contained inside it (when the input file is an archive, a
# document or some other kind of container).
# Value of 0 disables the limit.
# Note: disabling this limit or setting it too high may result in severe damage
# to the system.
# Technical design limitations prevent ClamAV from scanning files greater than
# 2 GB at this time.
# Default: 100M
#MaxFileSize 400M
```

---

### 7. Frontend: `dangerouslySetInnerHTML` Usage

**File:** One or more components in `app/src/`

At least one component uses `dangerouslySetInnerHTML` with what appears to be a hardcoded HTML string. While currently safe (static content), this pattern is fragile — if this value is ever sourced from an API response or user input in the future, it becomes an XSS vector.

**Recommendation:** Replace with a structured React component tree or use a sanitized renderer (e.g., DOMPurify) and add a comment explaining why raw HTML is intentional here.


**Fixed by @mobaetzel on 24.03.2026 08:52**

---

## 🟡 Medium Severity / Nice to Fix

### 8. Leaflet Map Component — Missing CSS Import Check

**Files:** `app/package.json` (+`leaflet`, `react-leaflet`), `app/src/components/map-point-field/`

Leaflet requires its CSS to be imported (`leaflet/dist/leaflet.css`) for the map to render correctly. If the CSS import is missing or only conditionally loaded, the map will render broken. Please verify the CSS import is in place globally or in the component.

**Fixed by @mobaetzel on 24.03.2026 09:02**

---

### 9. `createAvailableDataKey()` — Potential Performance Issue on High Collision

**File:** `src/main/java/de/aivot/GoverBackend/process/controllers/ProcessNodeController.java`

The import endpoint calls `createAvailableDataKey()` which iterates up to 9,999 times checking for key collisions in the database. In a process with many nodes, this generates up to 9,999 individual DB queries per import. 

**Recommendation:** Load all existing keys once (single query) and check against the in-memory set.

**Fixed by @mobaetzel on 24.03.2026 09:21**

---

### 10. Removed Undo Migrations — No Rollback Strategy Documented

**Files:** 20 `U*.sql` files deleted

Removing Flyway undo migrations is a deliberate and acceptable design decision for a major version (5.0.0), but the PR and CHANGELOG do not document:
- That rollback from this version requires manual DB intervention
- What the recommended rollback procedure is (snapshot restore, manual SQL, etc.)

**Recommendation:** Add a note to `CHANGELOG.md` and deployment docs clarifying the rollback posture for this release.

**Reviewed by @mobaetzel on 24.03.2026 08:54:** Currently irrelevant and needs no further actions.

---

### 11. `claude.yml` GitHub Actions Workflow — Uses Non-Existent `actions/checkout@v6`

**File:** `.github/workflows/claude.yml`

```yaml
uses: actions/checkout@v6
```

As of early 2026, `actions/checkout@v6` does not exist — the latest is `v4`. This will cause the Claude Code Action workflow to fail on every trigger.

**Action required:** Change to `actions/checkout@v4`.

**Reviewed by @mobaetzel on 24.03.2026 08:54:** This version exists. v6.0.2 was released on 09.01.2025

---

### 12. New Search View (`v_search_items`) Lacks `OR`-based Full-Text for Storage Providers

**File:** `src/main/resources/db/migration/V28_0_0__search_items.sql`

The `v_search_items` view uses `to_tsvector('german', ...)` for full-text search across all entity types, which is correct. However, storage providers appear to be added to the UNION but should be verified that the `key` column (which is an enum-like string, e.g. `"local_disk"`) is meaningful as a search term for end users. Consider supplementing with the `name` or `description` fields from the storage provider configuration JSON if available.

**Reviewed by @mobaetzel on 24.03.2026 00:10:** There is nothing to do here. The search condition for the name of the storage provider is sufficient and the storage provider definition key is no expressive enough for users to identify the storage provider further.

---

### 13. `ProcessNodeDefinitionKey` Max Length — Inconsistency Between Java and DB

**File:** `src/main/java/de/aivot/GoverBackend/process/entities/ProcessNodeEntity.java` + relevant migration

The Java `@Size(max = 128)` change should be reflected in a corresponding `ALTER TABLE` migration. Verify that a migration exists to extend the column from `VARCHAR(32)` to `VARCHAR(128)`. If no migration was added, inserts with keys between 33–128 chars will fail at the DB level.

**Checked by @mobaetzel on 24.03.2026 09:05:** Migration exists and has the correct length.

---

## 🟢 Positive Observations

- **Permission system decentralization** is a significant improvement. Moving from a single `Permissions.java` constant class to per-module `PermissionProvider` implementations makes permission discovery, documentation, and testing much easier.

- **Audit log builder pattern** (`auditService.create().withUser().withAuditAction().withDiff().log()`) is clean, expressive, and automatically captures IP addresses from `X-Forwarded-For`. This is a big UX improvement for operators.

- **Plugin architecture with semantic versioning** is well-designed. The `PluginComponent` interface with MAJOR.MINOR.PATCH versioning and explicit API stability guarantees will make long-term maintenance of custom nodes much more tractable.

- **`ElementData` → `AuthoredElementValues` + `DerivedRuntimeElementData` split** is architecturally sound. Separating user-authored input from derived/computed state eliminates an entire class of bugs where computed values were accidentally persisted or sent back to the server.

- **`ProcessNodeExportService.cleanConfigurationForExport()`** hook is a good security pattern — providers are responsible for sanitizing their own config before it leaves the system.

- **`install.sh`** as a one-shot installer is a great UX improvement for self-hosted deployments. The use of `openssl rand` for all secrets and `ON CONFLICT DO NOTHING` in migrations shows careful thought around idempotency.

- **Test coverage** for the new services is comprehensive: 331 lines for `ElementDerivationServiceV2`, 352 for `ApprovalActionNodeV1`, 340 for `DestinationKeyPayloadService`, plus security hardening tests.

- **`HEALTHCHECK`** instruction added to Dockerfile with mode-aware `healthcheck.sh` is a solid operational improvement.

- **OpenAPI docs disabled by default** (`GOVER_ENABLE_API_DOCS: false`) is a sensible security default.

---

## Summary Table

| # | Severity | Area | Finding |
|---|----------|------|---------|
| 1 | 🔴 Critical | Security | Export signature field is TODO |
| 2 | 🔴 Critical | DevOps | `install.sh` generates outdated Keycloak config |
| 3 | 🟠 High | Data Integrity | Validation message not updated after size change |
| 4 | 🟠 High | Audit | Diff calculation misses nested non-Map types |
| 5 | 🟠 High | Performance | Retention cleanup has no batch limit |
| 6 | 🟠 High | Operations | 5× upload size increase undocumented |
| 7 | 🟠 High | Security | `dangerouslySetInnerHTML` pattern is fragile |
| 8 | 🟡 Medium | Frontend | Leaflet CSS import unverified |
| 9 | 🟡 Medium | Performance | `createAvailableDataKey` N+1 queries |
| 10 | 🟡 Medium | Operations | Undo migration removal undocumented |
| 11 | 🟡 Medium | CI/CD | `actions/checkout@v6` doesn't exist |
| 12 | 🟡 Medium | Search | Storage provider search terms may not be meaningful |
| 13 | 🟡 Medium | Data Integrity | DB column width migration may be missing for key length change |

---

*Review generated by Claude Code — covers all 868 changed files across backend (Java/Spring Boot), frontend (React/TypeScript), database migrations, infrastructure, and CI/CD.*

mobaetzel:
Review findings from `feature/prototype-consolidation` -> `milestone/5.0.0` (asset module + connected storage provider changes):

1. High: deleting an asset folder removes asset rows from every storage provider that happens to share the same path prefix.
   In `AssetController.deleteFolder()` we already delete the storage-index subtree for the selected provider via `storageService.deleteFolder(storageProvider.getId(), folderPath)`, and the FK added in `V26_3_0__asset_storage_provider_references.sql` already cascades asset cleanup for that provider/path. The extra `assetRepository.deleteAllByStoragePathFromRootStartingWith(folderPath)` call is global and ignores `storageProviderId`, so deleting `/images/` from provider A will also delete asset rows under `/images/...` in provider B. See `src/main/java/de/aivot/GoverBackend/asset/controllers/AssetController.java:211-213` and `src/main/java/de/aivot/GoverBackend/asset/repositories/AssetRepository.java:10-13`.
  **Fixed by @mobaetzel on 2026-03-24 09:53**

2. High: S3 folder copy/move can delete the destination even when the source folder does not exist.
   `S3StorageProviderDefinitionV1.copyFolder()` deletes the target prefix before it has established that the source prefix contains any objects. If the source is missing or misspelled, `foundAtLeastOneObject` stays false, the method throws, and the previous destination content is already gone. `moveFolder()` inherits the same behavior because it delegates to `copyFolder()`. See `src/main/java/de/aivot/GoverBackend/plugins/core/v1/storage/S3StorageProviderDefinitionV1.java:328-329` and `:375-376`.
  **Fixed by @mobaetzel on 2026-03-24 10:10**

3. Medium: system storage providers are still mutable through the update endpoint.
   `StorageProviderService.performUpdate()` special-cases system providers, but when the request sets `status = SyncPending` it only enqueues a resync and then continues through the normal update flow. That means the same request can still overwrite name/description/max file size/read-only/test-provider/metadata/config on a provider that is supposed to be non-editable. This looks unintended given the explicit error text and delete protection. See `src/main/java/de/aivot/GoverBackend/storage/services/StorageProviderService.java:119-129` followed by `:144-158`.
  **Fixed by @mobaetzel on 2026-03-24 10:23**

danielkratz:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/aivot-digital/gover/pull/71#pullrequestreview-3991819950)

jp-performance (approved):

PR #87, opened 2026-04-24
refactor: standardize technical timestamps on timestamptz/Instant and centralize timezone handling

danielkratz:
This PR standardizes technical timestamps across the application on PostgreSQL `timestamptz` and Java `Instant`, while keeping only deputy validity windows as local wall-clock timestamps.

It also centralizes runtime timezone handling via `gover.timezone` / `GOVER_TIMEZONE`, removes hard-coded runtime timezone assumptions, and fixes temporal serialization inconsistencies that caused spurious diffs for `date` / `datetime` values in process action nodes.

## What changed

- Switched technical/system timestamps from `timestamp` / `LocalDateTime` to `timestamptz` / `Instant` across the main backend domains, including:
  - process definitions and process instances
  - forms and form revisions
  - audit logs
  - presets
  - storage
  - payments
  - departments, teams, roles, provider links
  - data objects
  - submissions and related DTOs / filters / projections

- Updated the existing Flyway create migrations so these columns are created as `timestamptz` directly.
- Kept deputy validity periods as local timestamps and made SQL-side comparisons explicit via `${gover_timezone}` instead of relying on the database session timezone.
- Added centralized timezone configuration with:
  - `gover.timezone`
  - `GOVER_TIMEZONE`
  - dynamic `TZ` / `-Duser.timezone` handling in the container entrypoint
- Replaced hard-coded runtime timezone usage in business logic, schedulers, formatting, and helper code with the centralized application timezone.
- Kept API serialization on ISO-8601 strings for `Instant` values.
- Normalized payload values before merge/diff so semantically equivalent `date` / `datetime` values no longer create false diffs in `DataChangeActionNode` / `ManualActionNode`.
- Hardened NoCode date parsing so invalid inputs surface as controlled `NoCodeException`s instead of uncaught runtime exceptions.
- Updated demo/example data and tests to match the new timestamp model.

## Review focus

Please review these areas especially carefully:

- Confirm that every field migrated to `Instant` is truly a technical/system timestamp and not a business-local wall-clock value.
- Review process business logic around deadlines, postponements, retries, retention, scheduler execution, and mail/PDF rendering, since these now resolve their zone from `gover.timezone`.
- Confirm that deputy validity logic is the only intended exception that remains local-time based.
- Review API / DTO / projection / filter changes for consumers that may previously have relied on naive local datetime strings.
- Review all SQL that still performs local-time comparisons and verify that `${gover_timezone}` is the correct business timezone abstraction.
- Review action-node payload handling for `date` / `datetime` fields, especially the new canonicalization before diff generation.
- Review NoCode and JS date helpers carefully, especially parsing / formatting assumptions and failure behavior for invalid ISO timestamps.
- Check whether any direct database queries, reports, scripts, or external integrations depended on the previous `timestamp without time zone` semantics.


src/main/resources/db/migration/V16_0_0__deputies.sql
```diff
@@ -27,8 +27,8 @@ with recursive deputy_hierarchy as (
            ud.deputy_user_id   as deputy_user_id,
            1                   as depth
     from user_deputies ud
-    where (ud.until_timestamp is null or ud.until_timestamp > now())
-      and ud.from_timestamp < now()
+    where (ud.until_timestamp is null or ud.until_timestamp > (now() at time zone '${gover_timezone}'))
+      and ud.from_timestamp < (now() at time zone '${gover_timezone}')
```
mobaetzel: This injects the set timezone, when the migration takes place, into this view. When the Timezone of the system changes, this view is not updated because the timezone is already set in the view and the migration ist done. Maybe we can use the timezone of the database and set it correctly instead of a one-time-injection here.

src/main/resources/db/migration/V1_0_0__initial.sql
```diff
@@ -5,8 +5,8 @@ CREATE TABLE system_configs
     key           VARCHAR(64) PRIMARY KEY,
     value         VARCHAR(96) NOT NULL,
     public_config BOOLEAN     NOT NULL DEFAULT false,
-    created       TIMESTAMP   NOT NULL,
-    updated       TIMESTAMP   NOT NULL
+    created       timestamptz NOT NULL,
+    updated       timestamptz NOT NULL
```
mobaetzel: Please do not change migrations, already applied in production.

src/main/resources/db/migration/V1_1_0__preset_versions.sql
```diff
@@ -22,11 +22,11 @@ CREATE TABLE preset_versions
     version VARCHAR(11) NOT NULL DEFAULT '1.0.0',
     root    jsonb       NOT NULL DEFAULT '{}',
 
-    published_at timestamp NULL,
-    published_store_at timestamp NULL,
+    published_at timestamptz NULL,
+    published_store_at timestamptz NULL,
```
mobaetzel: Please do not change migrations, already applied in production.

src/main/resources/db/migration/V2_0_0__keycloak_auth.sql
```diff
@@ -4,7 +4,7 @@ CREATE TABLE assets
 (
     key         uuid PRIMARY KEY,
     filename    VARCHAR(255) NOT NULL,
-    created     TIMESTAMP    NOT NULL,
+    created     timestamptz  NOT NULL,
```
mobaetzel: Please do not change migrations, already applied in production.

src/main/resources/db/migration/V2_7_0__destination_timestamp.sql
```diff
@@ -1,4 +1,4 @@
 -- alter existing table
 
 ALTER TABLE submissions
-    ADD COLUMN destination_timestamp TIMESTAMP;
\ No newline at end of file
+    ADD COLUMN destination_timestamp timestamptz;
```
mobaetzel: Please do not change migrations, already applied in production.

src/main/resources/db/migration/V2_11_0__form_revisions.sql
```diff
@@ -5,6 +5,6 @@ create table form_revisions
     id        bigserial primary key,
     form_id   integer      not null references forms (id),
     user_id   varchar(255) not null,
-    timestamp timestamp    not null,
+    timestamp timestamptz  not null,
```
mobaetzel: Please do not change migrations, already applied in production.

src/main/resources/db/migration/V3_2_2__payment_transaction.sql
```diff
@@ -8,6 +8,6 @@ create table payment_transactions
     payment_information  jsonb        null,
     payment_error        text         null,
     redirect_url         varchar(255) not null,
-    created              timestamp    not null,
-    updated              timestamp    not null
-)
\ No newline at end of file
+    created              timestamptz  not null,
+    updated              timestamptz  not null
```
mobaetzel: Please do not change migrations, already applied in production.

src/main/java/de/aivot/GoverBackend/utils/ApplicationTimeZone.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Nonnull
+    public static ZoneId getZoneId() {
+        return zoneId;
+    }
+
+    @Nonnull
+    public static String getZoneIdValue() {
+        return getZoneId().getId();
```
mobaetzel, line 22: Maybe we can pass this information into the frontend for the debug information. Use the AppConfigController for this if you implement it.

src/main/resources/db/migration/V17_5_0__user_full_name.sql
```diff
@@ -12,7 +12,9 @@ create view v_user_deputy_with_details as
 select udep.id                                                                                        as id,
        udep.from_timestamp                                                                            as from_timestamp,
        udep.until_timestamp                                                                           as until_timestamp,
-       udep.until_timestamp is null or (udep.from_timestamp < now() and udep.until_timestamp > now()) as active,
+       udep.until_timestamp is null or
+       (udep.from_timestamp < (now() at time zone '${gover_timezone}') and
+        udep.until_timestamp > (now() at time zone '${gover_timezone}'))                              as active,
```
mobaetzel: Here the same problems with the "hard" set timezones exists. This should be relative and not fixed during the migration. See the comment in the V16_0_0__deputies.sql.

docker-compose.yml
```diff
@@ -186,6 +186,7 @@ services:
       GOVER_ENVIRONMENT: "{{ SENTRY_ENVIRONMENT }}" # Setzen Sie die Umgebung für Sentry. Diese wird verwendet, um Fehlermeldungen in Sentry zu gruppieren. Passen Sie diese Umgebung unbedingt an. Entfernen Sie diesen Wert, um Sentry zu deaktivieren.
 
       GOVER_HOSTNAME: "https://{{ HOSTNAME }}" # Die Adresse inklusive Protokoll und Hostname unter der die Gover-Anwendung erreichbar ist. Passen Sie diese Adresse unbedingt an.
+      GOVER_TIMEZONE: "{{ GOVER_TIMEZONE }}" # Optional: Steuert die fachliche Zeitzone der App. Wenn die Zeile entfernt wird, verwendet Gover Europe/Berlin.
```
mobaetzel: Maybe we need to also set the timezone for the gover postgres database with the PGTZ env variable: https://www.postgresql.org/docs/current/libpq-envars.html

src/main/java/de/aivot/GoverBackend/elements/models/elements/form/input/DateInputElement.java
```diff
@@ -20,7 +22,11 @@
 
 public class DateInputElement extends BaseInputElement<ZonedDateTime> implements PrintableElement<ZonedDateTime> {
     private static final Logger logger = LoggerFactory.getLogger(DateInputElement.class);
-    public final static ZoneId zoneId = ZoneId.of("Europe/Berlin");
+
+    @Nonnull
+    public static ZoneId getZoneId() {
```
mobaetzel: We should remove this so that there is only one point to get the correct zone id from. If we cannot remove this here, we should at least mark it as deprecated.

src/main/java/de/aivot/GoverBackend/elements/models/elements/form/input/DateTimeInputElement.java
```diff
@@ -16,7 +17,9 @@
 import java.util.Objects;
 
 public class DateTimeInputElement extends BaseInputElement<ZonedDateTime> implements PrintableElement<ZonedDateTime> {
-    private static final ZoneId zoneId = ZoneId.of("Europe/Berlin");
+    private static ZoneId zoneId() {
+        return DateInputElement.getZoneId();
+    }
```
mobaetzel: The Date Range reuses the DateInputElement function for this. Remove this duplicate functionality.

src/main/java/de/aivot/GoverBackend/elements/models/elements/form/input/TimeInputElement.java
```diff
@@ -17,7 +18,10 @@
 import java.util.regex.Pattern;
 
 public class TimeInputElement extends BaseInputElement<ZonedDateTime> implements PrintableElement<ZonedDateTime> {
-    private final static ZoneId zoneId = ZoneId.of("Europe/Paris");
+    // Keep the standalone time element aligned with the rest of the office-time UI handling.
+    private static ZoneId zoneId() {
+        return DateInputElement.getZoneId();
+    }
```
mobaetzel: We should directly use the zoneId from the ApplicationTimeZone and not wrap it everywhere we use it.

src/main/java/de/aivot/GoverBackend/mail/services/ProcessTaskMailService.java
```diff
@@ -63,9 +64,9 @@ public void sendAssigned(@Nullable UserEntity triggeringUser,
                 .filter(value -> value != null && !value.isBlank())
                 .toList();
 
-        String startedLabel = processInstanceTask.getStarted().format(DATE_TIME_FORMATTER);
+        String startedLabel = DATE_TIME_FORMATTER.withZone(ApplicationTimeZone.getZoneId()).format(processInstanceTask.getStarted());
```
mobaetzel: You can directly initialize the DATE_TIME_FORMATTER with the correct timezone at the top of this class.

src/main/java/de/aivot/GoverBackend/ozgCloud/services/OZGCloudDataFormatService.java
```diff
@@ -366,7 +366,7 @@ private OZGCloudFormDataItem buildDateField(
                 .format(
                         DateTimeFormatter
                                 .ofPattern("yyyy-MM-dd")
-                                .withZone(DateInputElement.zoneId)
+                                .withZone(DateInputElement.getZoneId())
```
mobaetzel: Maybe we can initialize this formatter as a static final field of this class.

src/main/java/de/aivot/GoverBackend/plugins/core/v1/javascript/DateJavascriptV1.java
```diff
@@ -12,6 +13,10 @@
 
 @Service
 public class DateJavascriptV1 implements JavascriptFunctionProvider {
+    private static ZoneId zoneId() {
+        return ApplicationTimeZone.getZoneId();
+    }
+
```
mobaetzel: This should not wrap the ApplicationTimeZone

src/main/java/de/aivot/GoverBackend/plugins/core/v1/nodes/actions/ManualActionNodeV1.java
```diff
@@ -486,7 +486,7 @@ private ProcessNodeExecutionResultTaskCompleted completeTask(@Nonnull ProcessNod
         nodeData.put(OUTPUT_DIFF, diff);
         nodeData.put(OUTPUT_REMARK, remark);
         nodeData.put(OUTPUT_PROCESSED_BY_USER_ID, context.getCallingUser().getId());
-        nodeData.put(OUTPUT_PROCESSED_AT, LocalDateTime.now().toString());
+        nodeData.put(OUTPUT_PROCESSED_AT, Instant.now().toString());
```
mobaetzel: I think this dataset will be serialized so no toString() is required here

src/main/java/de/aivot/GoverBackend/plugins/core/v1/nodes/actions/DataChangeActionNodeV1.java
```diff
@@ -335,7 +336,7 @@ private ProcessNodeExecutionResultTaskCompleted completeTask(@Nonnull ProcessNod
         nodeData.put(OUTPUT_DIFF, diff);
         nodeData.put(OUTPUT_REMARK, remark);
         nodeData.put(OUTPUT_PROCESSED_BY_USER_ID, context.getCallingUser().getId());
-        nodeData.put(OUTPUT_PROCESSED_AT, LocalDateTime.now().toString());
+        nodeData.put(OUTPUT_PROCESSED_AT, Instant.now().toString());
```
mobaetzel: I think this will be serialized anyway so no toString() is needed here

src/main/java/de/aivot/GoverBackend/plugins/core/v1/nodes/actions/NoCodeActionNodeV1.java
```diff
@@ -678,6 +695,11 @@ private static String castToDateTime(Object value) throws ProcessNodeExecutionEx
         );
     }
 
+    @Nonnull
+    private static ZoneId zoneId() {
+        return ApplicationTimeZone.getZoneId();
+    }
+
```
mobaetzel: The ApplicationTimeZone should be used directly and not wrapped

mobaetzel (changes requested):

src/main/java/de/aivot/GoverBackend/utils/ApplicationTimeZone.java
danielkratz: That's a great suggestion and is now implemented.

mobaetzel (approved):

PR #90, opened 2026-05-11
fix: pin Traefik backend network on multi-homed routed services

dko-strd:
## Problem

The compose template multi-homes `gover` (gover + common) and `keycloak` (keycloak + common) without setting `traefik.docker.network`. When a routed container is attached to multiple Docker networks and the label is absent, Traefik picks one at random. If Traefik picks a network it isn't itself attached to (Traefik sits only on `common`), requests to that backend silently time out: TLS handshake succeeds, the HTTP request is forwarded toward an unreachable IP, no error in Traefik's own logs.

The selection is re-rolled on every Traefik config rebuild (startup + Docker container events), so the failure manifests sporadically across restarts and is easy to misdiagnose as a backend or DNS issue.

## Fix

- Set `traefik.docker.network=gover_common` on the two multi-homed routed backends (`keycloak`, `gover`).
- Pin the `common` network's name via the compose `name:` field, so the label value does not depend on the deployer's compose project name.

## Operator note

On upgrade, the previously project-prefixed `<project>_common` network becomes orphaned (zero containers). It can be cleaned up with `docker network rm <project>_common` after the next stack restart.


danielkratz:
@dko-strd Thank you for another contribution to Gover! 🙌

This is a very helpful fix, especially because Traefik’s random network selection in multi-network setups can be hard to reproduce and easy to misdiagnose.

We really appreciate your continued support!

danielkratz (approved):

PR #92, opened 2026-05-20
devops: update versions of services and requirements

mobaetzel:
Update the versions of some dependent services and of the min requirements for a deployment.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,7 +50,7 @@ services:
 
 
   redis:
-    image: redis:7.4.14
+    image: redis:7.4.9
     restart: always
     networks:
       - gover
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
 
 
   gover_database:
-    image: postgres:15.15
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -43,14 +43,14 @@ services:
...
-    image: clamav/clamav-debian:1.5.1
+    image: clamav/clamav-debian:1.5.2
     restart: always
     networks:
       - gover
 
 
   redis:
-    image: redis:7.4.7
+    image: redis:7.4.14
```
danielkratz: The latest Redis image on Docker Hub is 7.4.9, which appears to also be the latest release from two weeks ago. 7.4.14 does not seem to be a real version (at least for the open source branch). Please check and possibly fix this. Thanks!

danielkratz (changes requested):

docker-compose.yml
mobaetzel: Yo are right, i've looked for Redis 7.2.X. Thanks for catching this!

danielkratz (approved):

PR #108, opened 2026-07-07
fix: process publish check intermediate fix

mobaetzel:
# Description
The current permission check for publishing process versions does not check the system permissions. This PR fixes this issue.

# Ticket
- OP#885


mobaetzel:
This will be solved on the feature branch `feature/permission-check-basis`

PR #109, opened 2026-07-10
feat: add basics for resource-aware permission enforcement and staff UI handling

danielkratz:
This PR introduces the basics for resource-aware permission handling across backend and staff frontend.

### Backend

- Replaces the former permission scope model with `supportsDomainRoleAssignment`
- Adds system-wide and resource-specific permission checks for departments, teams, processes, process instances, roles, themes, secrets, audit logs, and related resources
- Extends the permission set API to include department, team, process, process instance, domain, and system permissions
- Adds a dedicated permission for reading permission sets
- Validates system and domain role permissions on save
- Keeps already assigned unavailable permissions editable so roles can be cleaned up safely

### Frontend

- Loads permission sets during staff application startup
- Adds frontend permission utilities and hooks for `check...` and `has...` access checks
- Synchronizes permission refreshes across browser tabs
- Refreshes permissions after access-changing actions such as role updates, membership role changes, process creation, and process permission changes
- Hides staff navigation entries when the user lacks the required read permission
- Shows StaffShell access-denied errors for direct route access without read permission
- Adds shared UI primitives for disabled actions with permission tooltips
- Applies permission-aware list and detail behavior to departments, teams, system roles, domain roles, themes, secrets, and audit logs

OP#836
OP#837


backend/src/main/java/de/aivot/gover/backend/department/services/DepartmentService.java
```diff
@@ -147,13 +154,73 @@ private void validateRequiredStringSetting(
     ) throws ResponseException {
         if (value == null) {
             if (isRoot) {
-                throw ResponseException.badRequest("Für Organisationseinheiten der obersten Ebene muss „%s“ konfiguriert sein.", fieldName);
+                throw ResponseException.badRequest(String.format(
+                        "Für Organisationseinheiten der obersten Ebene muss „%s“ konfiguriert sein.",
+                        fieldName
+                ));
             }
```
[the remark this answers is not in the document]
danielkratz: Not applying this suggestion. `ResponseException.badRequest` also has a `(String message, String details)` overload, so passing a `String` as the second argument does not select the varargs formatter. The explicit `String.format(...)` is required here; otherwise `%s` remains unformatted, which is covered by `DepartmentServiceTest`.

app/src/modules/teams/pages/details/teams-details-page.tsx
```diff
@@ -6,15 +6,53 @@ import {GenericPageHeaderPropsHelpDialog} from '../../../../components/generic-p
...
+
     return (
         <PageWrapper
             title="Team bearbeiten"
             fullWidth
             background
         >
             <GenericDetailsPage<TeamEntity, number, void>
+                hasAccess={hasAccess}
+                isEditable={isEditable}
```
mobaetzel: Maybe it's a good idea to integrate the permission checking directly into the generic details page instead of always supplying these checks but keeping these check for non-standard access and edit check behaviour. i boils down to a system permission check for a given permission for new entries, a team check and a department specific check for non new entries additionally. We can add these as fixed props and reduce the repetition in the components reusing the generic details page.

app/src/modules/teams/pages/details/teams-details-page.tsx
```diff
@@ -28,7 +66,10 @@ export function TeamsDetailsPage() {
                     {
                         path: '/teams/:id/members',
                         label: 'Teammitglieder',
-                        isDisabled: (item) => !item?.id,
+                        isDisabled: (item) => !item?.id || !checkTeamPermission(permissions, item.id, Permission.TEAM_MEMBERSHIP_READ),
+                        disabledTooltip: (item) => item?.id
+                            ? formatMissingPermissionTooltip(Permission.TEAM_MEMBERSHIP_READ)
+                            : undefined,
```
mobaetzel: When upgrading the props of the generic details page we can also extend the tags with a specific permission check as well as a `onlyExisting` flag. The generic details page knows already about the new/exisiting state so this might be a simplification of the tabs

app/src/modules/teams/pages/details/teams-details-page-index.tsx
```diff
@@ -71,29 +51,29 @@ export function TeamsDetailsPageIndex() {
 
     const apiService = useMemo(() => new TeamsApiService(), []);
     const changeBlocker = useChangeBlocker(item, team);
+    const editPermission = team?.id === 0 ? Permission.TEAM_CREATE : Permission.TEAM_UPDATE;
```
mobaetzel: The GenericDetailsPageContext exposes the isNewItem-Flag which should be used if possible.

backend/src/main/java/de/aivot/gover/backend/process/filters/ProcessAccessControlFilter.java
```diff
@@ -6,10 +6,13 @@
 import jakarta.annotation.Nonnull;
 import org.springframework.data.jpa.domain.Specification;
 
+import java.util.List;
+
 public class ProcessAccessControlFilter implements Filter<ProcessAccessControlEntity> {
     private Integer sourceTeamId;
     private Integer sourceDepartmentId;
     private Integer targetProcessId;
+    private List<Integer> targetProcessIds;
```
mobaetzel, line 15: These filter fields have been added to the backend filters but not the frontend filter interfaces. Was this intentionally?

app/src/modules/teams/pages/details/teams-details-page-index.tsx
danielkratz: I've updated all places in the example resources accordingly.

backend/src/main/java/de/aivot/gover/backend/process/filters/ProcessAccessControlFilter.java
danielkratz: Nope, that's an oversight. I added this field to the two frontend interfaces.

app/src/modules/teams/pages/details/teams-details-page.tsx
danielkratz: I have initially decided against it and not revisited the topic yet. After careful consideration I agree with the idea and implemented this as explained by you.

app/src/modules/teams/pages/details/teams-details-page.tsx
danielkratz: This has also been implemented.

backend/src/main/java/de/aivot/gover/backend/search/services/SearchRecentItemService.java
```diff
@@ -0,0 +1,139 @@
...
+                ? retentionDays
+                : DEFAULT_RETENTION_DAYS;
+    }
+
+    @Nonnull
+    public List<SearchItemResponseDTO> listVisibleRecentItems(
+            @Nonnull String userId,
+            int size
+    ) {
+        var normalizedSize = Math.max(1, Math.min(size, maxItemsPerUser));
```
mobaetzel: This can be simplified with `Math.clamp(...)`.

backend/src/main/java/de/aivot/gover/backend/search/services/SearchRecentItemService.java
```diff
@@ -0,0 +1,139 @@
...
+        searchRecentItemRepository.upsert(
+                userId,
+                request.originTable(),
+                request.id()
+        );
+        searchRecentItemRepository.deleteOverflow(userId, maxItemsPerUser);
+    }
+
+    @EventListener(ApplicationReadyEvent.class)
+    public void deleteExpiredRecentItemsOnStartup() {
```
mobaetzel, line 99: This method calls a Transactional and therefore should also be marked as `@Transactional`.

backend/src/main/java/de/aivot/gover/backend/search/services/SearchRecentItemService.java
```diff
@@ -0,0 +1,139 @@
...
+        deleteExpiredRecentItems("startup");
+    }
+
+    @Scheduled(cron = "0 30 3 * * *", zone = "${gover.timezone}")
+    public void deleteExpiredRecentItemsNightly() {
```
mobaetzel, line 105: This method calls a Transactional and therefore should also be marked as `@Transactional`.

backend/src/main/java/de/aivot/gover/backend/mail/controllers/TestMailController.java
```diff
@@ -33,29 +35,33 @@
...
     ) throws ResponseException {
         var user = userService
                 .fromJWT(jwt)
-                .orElseThrow(ResponseException::unauthorized)
-                .asSuperAdmin()
-                .orElseThrow(ResponseException::forbidden);
+                .orElseThrow(ResponseException::unauthorized);
+
+        permissionService
+                .requireSystemPermission(user.getId(), ConfigPermissionProvider.SYSTEM_CONFIG_UPDATE);
```
mobaetzel, line 65: Why is this endpoint bound to the `SYSTEM_CONFIG_UPDATE` specifically?

backend/src/main/java/de/aivot/gover/backend/permissions/controllers/PermissionsController.java
```diff
@@ -87,14 +136,169 @@ public PermissionSet listForUser(
...
+        return permission != null && permission.getDepartmentId() != null && hasPermissionEntries(permission.getPermissions());
+    }
+
+    private static boolean hasPermissionEntries(@Nullable List<String> permissions) {
+        return permissions != null && !permissions.isEmpty();
+    }
+
+    private List<DomainPermission> listDomainPermissions(@Nonnull String userId) {
+        // The view may contain structural rows without effective grants, so the API exposes only non-empty permission arrays.
+        return jdbcTemplate.query(
```
mobaetzel: This seems to only fetch a subset of columns with some filtering. Instead of building this query maybe a projection of the entity can be used to select just the subset and perform the filtering. Interactions with the database (sql code etc.) should be done in the repositories if possible. This comment also applies to the other queries in this class. More information on projections can be found here: https://docs.spring.io/spring-data/jpa/reference/repositories/projections.html#projections.interfaces

backend/src/main/resources/db/migration/V29_5_0__search_items.sql
```diff
@@ -54,8 +65,28 @@ SELECT text 'teams'                AS origin_table,
...
+       name                        AS search_text,
+       utp.user_id                 AS user_id,
+       utp.permissions             AS permissions
 FROM teams
+         JOIN v_user_team_permissions AS utp
+              ON utp.team_id = teams.id
+
+UNION ALL
+
+-- Users
```
mobaetzel, line 77: Why does a normal user, who has no business in managing users, the ability to search for other users and, i presume, open their user page via the search dialogue? The search ability is bound to the USER_READ permission therefore are all users able to find other users in the search dialogue, ain't they?

backend/src/main/resources/application.yml
```diff
@@ -189,6 +189,10 @@ gover:
   departmentLevelLabels: "${GOVER_DEPARTMENT_LEVEL_LABELS:Organisation,Bereich,Abteilung,Sachgebiet,Arbeitsbereich}"
   security:
     csrfTtlSeconds: "${GOVER_CSRF_TTL_SECONDS:86400}"
+  search:
```
mobaetzel, line 192: These config properties are not bound to any class but access via dependency injection. we should try to always embody these configuration properties in classes to have them modelled in the OOP structure of our project.

backend/src/main/resources/db/migration/V29_5_0__search_items.sql
```diff
@@ -66,8 +97,12 @@ SELECT text 'process_nodes'                                         AS origin_ta
        id::varchar                                                  AS id,
        to_tsvector('german', coalesce(name, '')) ||
        to_tsvector('german', coalesce(description, ''))             AS searchable_element,
-       trim(coalesce(name, '') || ' ' || coalesce(description, '')) AS search_text
+       trim(coalesce(name, '') || ' ' || coalesce(description, '')) AS search_text,
+       upp.user_id                                                  AS user_id,
+       upp.permissions                                              AS permissions
 FROM process_nodes
+         JOIN v_user_process_access_permissions AS upp
```
mobaetzel, line 104: This view can yield the same process node and user id multiple times because the underlying `v_user_process_access_permissions` can produce a combination of user and process via a department and via a team. this needs to be aggregated and the permissions need to be combined via a distinct union.

backend/src/main/resources/db/migration/V29_5_0__search_items.sql
```diff
@@ -178,9 +254,12 @@ SELECT text 'processes'                                              AS origin_t
        p.internal_title || ' (Version ' || pv.process_version || ')' AS label,
        p.id::varchar || ',' || pv.process_version                    AS id,
        to_tsvector('german', p.internal_title)                       AS searchable_element,
-       p.internal_title                                              AS search_text
+       p.internal_title                                              AS search_text,
+       usp.user_id                                                   as user_id,
+       usp.permissions                                               as permissions
 FROM process_versions pv
          JOIN processes p ON pv.process_id = p.id
+         JOIN v_user_process_access_permissions AS usp ON usp.target_process_id = p.id -- TODO: Cross Join w/ v_user_system_permission
```
mobaetzel: This view can yield the same process node and user id multiple times because the underlying `v_user_process_access_permissions` can produce a combination of user and process via a department and via a team. this needs to be aggregated and the permissions need to be combined via a distinct union.

mobaetzel (changes requested):
I've found some minor things which i would like you to review. Additionally, I've noticed, that we now barely use the deprecated and dagerous `getIsSuperAdmin` and `getIsSystemAdmin` of the `UserEntity`. Maybe now is the right point in time to remove them finally.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessVersionController.java
```diff
@@ -189,14 +225,10 @@ public ProcessVersionEntity update(
                 .orElseThrow(ResponseException::notFound);
         var existingMap = AuditLogPayload.toMap(existing);
 
-        var processDefinition = processDefinitionService
-                .retrieve(existing.getProcessId())
-                .orElseThrow(ResponseException::badRequest);
-
         permissionService
```
mobaetzel, line 238: Here only update is required but via the DTO the status of the ProcessVersion can be updated in the ProcessVersionService bypassing the PUBLISH_LOCAL permission. The status needs to be overwritten in the dto based on the existing entity.

mobaetzel: THe same applies to the create endpoint of this controller.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessNodeController.java
```diff
@@ -191,6 +241,12 @@ public ProcessNodeEntity update(
                 .retrieve(id)
                 .orElseThrow(ResponseException::notFound);
 
+        permissionService.requireProcessPermission(
+                execUser.getId(),
+                existing.getProcessId(),
+                ProcessPermissionProvider.PROCESS_DEFINITION_UPDATE
+        );
+
         if (processNodeRepository.existsByDataKeyAndIdIsNotAndProcessIdAndProcessVersion(
```
mobaetzel, line 259: The updatedDTO needs to be injected with the already validated process id and version because in the service, the data from the dto is applied. With this a user can move a node out of a scope he/she has permission for. If the updatedDTO is not fixed with the already validated process id and version, the values from the dto are passed to the service and updated.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessEdgeController.java
```diff
@@ -127,6 +182,12 @@ public ProcessEdgeEntity update(
                 .retrieve(id)
                 .orElseThrow(ResponseException::notFound);
 
+        permissionService.requireProcessPermission(
```
mobaetzel, line 189: The updatedDTO needs to be injected with the already validated process id and version because in the service, the data from the dto is applied. With this a user can move a node out of a scope he/she has permission for. If the updatedDTO is not fixed with the already validated process id and version, the values from the dto are passed to the service and updated.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessInstanceTaskController.java
```diff
@@ -117,20 +145,11 @@ public ProcessInstanceTaskEntity create(
...
-                .retrieve(processDefinition.getDepartmentId())
-                .orElseThrow(ResponseException::badRequest);
-
-        permissionService
-                .testDepartmentPermission(
-                        execUser.getId(),
-                        department.getId(),
-                        ProcessPermissionProvider.PROCESS_DEFINITION_CREATE
-                );
+        permissionService.requireProcessInstancePermission(
```
mobaetzel: This method does not check if the node, process and process-version the task is created for belongs to the checked process instance. Maybe a api creation of task shouldn't even be possible.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessInstanceController.java
```diff
@@ -173,6 +199,12 @@ public ProcessInstanceEntity update(
                 .retrieve(id)
                 .orElseThrow(ResponseException::notFound);
 
+        permissionService.requireProcessInstancePermission(
```
mobaetzel, line 162: The service does not correctly check if the status, process and process-version updated dto are reachable for the user. Maybe process instances shouldn't be creatable or updatetable?

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessInstanceAccessControlController.java
```diff
@@ -77,6 +92,79 @@ protected Integer getIdForEntity(ProcessInstanceAccessControlEntity entity) {
...
+                permissionService.requireProcessInstancePermission(
+                        user.getId(),
+                        filter.getTargetProcessInstanceId().longValue(),
+                        ProcessPermissionProvider.PROCESS_INSTANCE_UPDATE
+                );
+            } else {
+                var accessibleProcessInstanceIds = permissionService
+                        .getProcessInstancesWithPermission(user.getId(), ProcessPermissionProvider.PROCESS_INSTANCE_UPDATE)
+                        .stream()
+                        .map(Long::intValue)
```
mobaetzel: This converts longs to ints and may produce an overflow when the int limit is reached. Why is this done here?

mobaetzel (changes requested):

backend/src/main/java/de/aivot/gover/backend/mail/controllers/TestMailController.java
danielkratz: We felt like this check should not be used by anybody and since this will probably get part of the settings instead of staying a standalone page we bound it to SYSTEM_CONFIG_UPDATE. Please let me know if you have a better suggestion or resolve this conversation. Thanks!

backend/src/main/resources/db/migration/V29_5_0__search_items.sql
danielkratz: We discussed this topic in the team and decided that there are valid use cases for the users to be included as a resource in the spotlight search.

backend/src/main/java/de/aivot/gover/backend/permissions/controllers/PermissionsController.java
danielkratz: Thanks, agreed. I refactored all three controller-local JDBC queries.

The filtering and column selection now live in repository query methods using closed interface projections. PermissionsController only orchestrates the repositories and maps the projection results to the existing API records.

For the process and process-instance permission views, the required JPA repository/entity mappings did not exist yet, so I added them as part of the refactoring.

The department and team repositories already apply the same filtering, so the redundant controller-side filters were removed as well.

backend/src/main/java/de/aivot/gover/backend/mail/controllers/TestMailController.java
danielkratz: This was resolved in person. We keep the binding to SYSTEM_CONFIG_UPDATE.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessInstanceTaskController.java
danielkratz: See separate comment with explanation.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessInstanceController.java
danielkratz: See separate comment with explanation.

backend/src/main/java/de/aivot/gover/backend/process/controllers/ProcessInstanceAccessControlController.java
danielkratz: Good catch. The conversion was compensating for an inconsistent type declaration. process_instances.id is defined as BIGSERIAL and represented as Long, while process_instance_access_controls.target_process_instance_id, its entity field and its filter were still using INTEGER/Integer. I therefore changed the foreign-key column to BIGINT, aligned the entity and filter with Long, and removed the narrowing conversion.

danielkratz:
@mobaetzel 

Thanks for the review. You were right about `getIsSuperAdmin()` and `getIsSystemAdmin()`: both methods have now been removed from `UserEntity`, together with the deprecated `asSuperAdmin()` and `asSystemAdmin()` helpers. Their remaining usages were replaced with explicit permission checks. The obsolete admin properties and related utility code were also removed from the frontend.

While reviewing the runtime resources based on your comments, I found that their generic CRUD endpoints allowed clients to modify engine-owned state directly. I therefore restricted the API as follows:

- Process instances can no longer be created or updated through generic `POST`/`PUT` endpoints. Creation remains part of the trigger flows, while legitimate mutations use explicit commands such as reassigning or restarting failed instances.
- Process instance tasks no longer expose generic create, update, or delete operations. User interaction is performed through the task contexts, with only the explicit failed-task rerun command remaining.
- Process instance events are now read-only because they form an append-only history maintained by the backend.
- Process instance attachments are now read-only. They are created through validated form/task execution flows and share the lifecycle of their process instance, so deleting the complete process instance is the only supported deletion path.
- Attachment sets remain read-only, and the previously missing process-instance permission checks were added.

The frontend API abstraction was adjusted accordingly: the read functionality was extracted from `BaseCrudApiService` into a new `BaseReadApiService`. Runtime resource services now extend the read-only base and expose only explicitly supported domain commands. Comments in the corresponding controllers document why generic mutation endpoints are intentionally absent.

Independently of your findings, I also noticed that several OpenAPI operation descriptions still referred to the old super-admin/system-admin model. I standardized the authorization descriptions across the backend so that they now reference the exact permission constants and explicitly distinguish system-level permissions, resource-scoped permissions, AND/OR requirements, self-access exceptions, and permission-filtered results. This also uncovered and corrected a few descriptions that did not match the actual permission checks. The remaining unused role-based `ResponseException` helpers were removed as well.

backend/src/main/resources/db/migration/V29_5_0__search_items.sql
```diff
@@ -66,10 +97,20 @@ SELECT text 'process_nodes'                                         AS origin_ta
        id::varchar                                                  AS id,
        to_tsvector('german', coalesce(name, '')) ||
        to_tsvector('german', coalesce(description, ''))             AS searchable_element,
-       trim(coalesce(name, '') || ' ' || coalesce(description, '')) AS search_text
+       trim(coalesce(name, '') || ' ' || coalesce(description, '')) AS search_text,
+       upp.user_id                                                  AS user_id,
+       array_unique_union_agg(upp.permissions)                      AS permissions
 FROM process_nodes
+         JOIN v_user_process_access_permissions AS upp
+              ON upp.target_process_id = process_nodes.process_id -- TODO: Cross Join w/ v_user_system_permission
```
mobaetzel: One minor thing and totally my fault: These TODOs for cross joins are outdated because the underlying views should already do the joining with all possible users and what not.

mobaetzel (approved):

jp-performance (approved):

PR #111, opened 2026-07-13
feat: write storage item node

mobaetzel:
# Description
This pull request introduces the automated process element "Write document to storage provider," enabling modelers to export and archive documents to external systems outside of Gover. The implementation includes configurable attributes for selecting the storage provider and specifying the file and target path. By utilizing this generic storage provider interface, we establish a unified approach that eliminates the future need to develop and maintain individual process elements for specific e-file systems.

# Ticket
- OP#270


backend/src/main/java/de/aivot/gover/backend/plugins/core/v1/nodes/actions/StoreAttachmentSetActionNodeV1.java
danielkratz, line 1: This node might be named in a way that is semantically suboptimal. Ultimately, the node is about storing data on an external storage provider. However, the emphasis of the name is on the storage of attachment sets. This somewhat misrepresents the intent.

backend/src/main/java/de/aivot/gover/backend/plugins/core/v1/nodes/actions/StoreAttachmentSetActionNodeV1.java
```diff
@@ -0,0 +1,1047 @@
...
+    public static class AttachmentSetStorageConfig {
+        public static final String ATTACHMENT_SET_DATA_KEYS_FIELD_ID = "attachment_set_data_keys";
+        public static final String STORAGE_PATH_FIELD_ID = "storage_path";
+        public static final String IGNORE_EMPTY_ATTACHMENT_SET_FIELD_ID = "ignore_empty_attachment_set";
+        public static final String CUSTOMIZE_FILE_NAME_FIELD_ID = "customize_file_name";
+        public static final String FILE_NAME_FIELD_ID = "file_name";
+
+        @InputElementPOJOBinding(id = ATTACHMENT_SET_DATA_KEYS_FIELD_ID, type = ElementType.ProcessInstanceAttachmentSetSelect, properties = {
+                @ElementPOJOBindingProperty(key = "label", strValue = "Anlagensatz"),
+                @ElementPOJOBindingProperty(key = "hint", strValue = "Anlagensatz der Prozessinstanz, dessen Anhänge gespeichert werden sollen."),
```
danielkratz, line 1004: As far as I know, when communicating with users, we don't use the term “Prozessinstanz” but rather “Vorgang.” I'll create an issue to have this updated in the few places where it's currently written that way.

backend/src/main/java/de/aivot/gover/backend/plugins/form/v1/nodes/FormTriggerNodeV1.java
```diff
@@ -32,32 +41,49 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
 import java.util.*;
 
 @Component
 public class FormTriggerNodeV1 implements ProcessNodeDefinition<FormTriggerConfigV1>, PluginComponent {
     public static final String NODE_KEY = "form";
     private static final String PORT_NAME = "input";
     private static final String COPY_VALUE_TEMPLATE_PATH_SEGMENT = "__copy_value__";
+    private static final String CUSTOMER_SUMMARY_FILE_NAME = "Formularausdruck.pdf";
```
danielkratz, line 63: Could we rename this file to `Formularzusammenfassung.pdf`? `Formularausdruck` is slightly misleading, as the PDF contains a summary of the submitted form data rather than a printout. The proposed name also matches the terminology already used throughout the code.

danielkratz (approved):

PR #114, opened 2026-07-15
feat: module limits

mobaetzel:
# Ticket
- OP#200
- OP#234
- OP#237


app/src/modules/process/pages/details/process-details-page.tsx
```diff
@@ -85,10 +85,19 @@ import {useProcessExport} from '../../../../hooks/use-process-export';
 import {ProcessVersionsDialog} from '../../dialogs/process-versions-dialog';
 import {NodeProblemsAlert} from '../../components/node-problems-alert';
 import {ProcessPublishDialog} from '../../dialogs/process-publish-dialog';
+import {getProcessNodeLimit, isFormModuleEnabled, isProcessNodeTypeUnlimited} from '../../../../utils/module-flags';
 
 export const SHOW_ERRORS_ROUTER_STATE = 'show-errors-on-load';
 
+const FORM_PLUGIN_KEY = 'de.aivot.form';
 const PROCESS_DETAILS_PAGE_SKELETON_DELAY = 250;
+// Obsolet, use ProviderTypeStyles instead
```
danielkratz: This comment mentions implicitly that PROCESS_NODE_TYPE_LABELS should not be used anymore. Just form them to be used in new code a 100 rows further down? Maybe clarify this comment (which includes a typo btw) or update the usage.

app/src/shells/customer/customer-shell-router.tsx
danielkratz, line 1: As we discussed in the team disabling the forms module will not automatically result in no homepage being shown or a 404 as the same page will be the entrypoint for the portals. And even if neither Portals or Forms are active we should show a page explaining that there is a Gover instance under this domain. This is a UX best practice so that user who mistakenly get directed to the root domain are not confused with an error message. So since this would be temporary I'm not sure if it's worth to add in the first place.

backend/src/main/java/de/aivot/gover/backend/plugins/form/FormModuleWebMvcConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    private record FormModuleInterceptor(@Nonnull GoverConfig goverConfig) implements HandlerInterceptor {
+        @Override
+        public boolean preHandle(@Nonnull HttpServletRequest request,
+                                 @Nonnull HttpServletResponse response,
+                                 @Nonnull Object handler) throws Exception {
+            if (!goverConfig.isFormModuleEnabled()) {
+                // Disabled public form URLs should look absent so external callers cannot depend on a hidden module.
+                throw ResponseException.notFound();
```
danielkratz, line 39: As we discussed in the team disabling the forms module will not automatically result in no homepage being shown or a 404 as the same page will be the entrypoint for the portals. And even if neither Portals or Forms are active we should show a page explaining that there is a Gover instance under this domain. This is a UX best practice so that user who mistakenly get directed to the root domain are not confused with an error message. So since this would be temporary I'm not sure if it's worth to add in the first place.

backend/src/main/resources/application.yml
```diff
@@ -213,6 +213,12 @@ gover:
   maxSubmissionCopyRetryCount: 3
   bootstrapAdminMail: "${GOVER_BOOTSTRAP_ADMIN_MAIL:}"
   registryHostname: "${GOVER_TEMPLATE_REGISTRY_URL:}"
+  moduleFlags: "${GOVER_MODULE_FLAGS:form,process,portal}"
+  processNodeLimits:
```
danielkratz, line 207: This suggests the process basic limits are the default? If I remember correctly we talked about the default limits and preferred the default to be unlimited. The background being that the limits only target a small part of the user base and the process basic limits as a default would not match the moduleFlags defaults.

danielkratz (changes requested):

app/src/modules/process/pages/details/process-details-page.tsx
mobaetzel: My intention was to signal that we should change this in the future because it is not mission critical right now.

backend/src/main/java/de/aivot/gover/backend/plugins/form/FormModuleWebMvcConfiguration.java
mobaetzel: I've added a new index page for this use case

backend/src/main/resources/application.yml
mobaetzel: As long as the process module flag is the default of the flags, these limits never take effect. i've already configured them to make future deployments easier.

danielkratz (approved):

danielkratz (approved):

PR #127, opened 2026-07-28
chore: upgrade frontend Vite toolchain to Vite 8

danielkratz:
Upgrades the frontend Vite toolchain from Vite 7 to Vite 8 while keeping TypeScript on 5.8.x.

This moves the production build onto Vite 8’s Rolldown-based pipeline and updates the related Vite plugins.

## Changes

- Upgraded:
  - `vite` from `7.3.6` to `8.1.5`
  - `@vitejs/plugin-react` from `4.x` to `6.x`
  - `vite-plugin-svgr` from `4.x` to `5.x`
  - `vite-plugin-checker` from `0.10.x` to `0.11.x`
- Kept `typescript` unchanged at `5.8.x`.
- Switched TypeScript module resolution from `node` to `bundler`.
- Simplified the React plugin config in the Vite and Vitest configs by removing the old Emotion/Babel options.
- Updated Monaco worker imports to use explicit `.js` file paths compatible with Rolldown resolution.
- Replaced private/deep type imports from MUI packages with public package exports.
- Removed local Altcha type declarations that now conflict with the package-provided types.
- Declared required peer dependencies explicitly:
  - `monaco-editor`
  - `yjs`
  - `@testing-library/dom`
  - `@typescript-eslint/parser`
- Added an override for `@napi-rs/wasm-runtime@1.1.6` to keep `npm ci` stable with Rolldown’s optional WASM dependency tree.

## Notes

This does not migrate the project to TypeScript 7. Vite 8’s Rust/Rolldown/Oxc-based improvements affect bundling and transpilation, while type checking still runs through the existing TypeScript 5.8 setup.

`moduleResolution: "bundler"` is intentional for Vite compatibility. It surfaced a few private package imports, which were replaced with public exports or explicit file paths.

## Verification

- `npm ci --dry-run`
- `npm run typecheck`
- `npm run test`
- `npm run build:customer`
- `npm run build:staff`


danielkratz:
After reviewing the browser compatibility implications of the Vite 8 upgrade and aligning with Product Management, I added the build target explicitly to both Vite configurations.

Our publicly communicated compatibility remains the latest three versions of Chrome, Edge, Firefox, and Safari. Technically, the production builds use Vite’s `baseline-widely-available` target. Making this explicit documents that distinction and highlights that the resolved Baseline may change with future Vite upgrades.

I also removed the existing Browserslist configuration because it is currently not consumed by any tool and does not control Vite’s build output. With the current Vite version, these changes do not alter the generated output; they clarify the intended configuration and prevent conflicting browser-support definitions.

mobaetzel (approved):

mobaetzel (approved):

PR #128, opened 2026-08-03
feature: standardize temporal data and application timezone handling

danielkratz:
This PR standardizes how Gover represents, transports, stores, and displays temporal values.

It removes implicit browser, JVM, and database timezone assumptions and establishes explicit contracts for calendar dates, local times, and absolute instants.

## Changes

- Introduce shared temporal utilities and a configurable business clock.
- Require absolute timestamps to contain `Z` or an explicit numeric offset.
- Serialize instants using the offset of the configured application timezone.
- Preserve empty timestamp strings as `null` for compatibility with existing form payloads.
- Represent calendar dates and local times without artificial timezone information.
- Replace frontend `Date` handling with typed ISO values and Luxon-based MUI pickers.
- Resolve and display absolute instants consistently in the application timezone.
- Add explicit Date, Time, and DateTime support to Low Code and NoCode APIs.
- Add temporal creation, formatting, extraction, combination, and comparison operators.
- Model deputy assignments as inclusive calendar-date ranges.
- Align mail, PDF, development example, and destination payload formatting with the new contracts.
- Add coverage for timezone offsets, DST gaps and overlaps, precision, serialization, validation, and deputy ranges.

OP#990


backend/src/main/java/de/aivot/gover/backend/plugins/core/v1/nodes/actions/ApprovalActionNodeV1.java
```diff
@@ -393,7 +394,10 @@ public Optional<ProcessNodeExecutionResult> onEventFromStaffTaskView(@Nonnull Pr
         nodeData.put(OUTPUT_DECISION, decision);
         nodeData.put(OUTPUT_REMARK, remarkText);
         nodeData.put(OUTPUT_PROCESSED_BY_USER_ID, context.getCallingUser().getId());
-        nodeData.put(OUTPUT_PROCESSED_AT, IsoTimestampUtils.nowUtc());
+        nodeData.put(
+                OUTPUT_PROCESSED_AT,
+                IsoTimestampUtils.nowWithOffset(ApplicationTimeZone.getZoneId())
+        );
```
mobaetzel: The nodeData is converted via the instance of the `ObjectMapperFactory` inside the `JsonObjectConverter`, which is the registered converter for the `nodeData` inside the `ProcessInstanceTaskView`. The `objectMapper` produced by the `ObjectMapperFactory` utilizes the `ApplicationTimeZoneInstantSerializer` to serialize `Instant.class`. Is this creation of a string-iso-timestamp when writing Instants to later-serialized-JSON-data required? It should be sufficient, to store an Instant in a dataset and then let the `JsonObjectConverter` respectively the instance of the `ObjectMapperFactory` to do its thing, doesn't it? This is not specific to these lines of code, more a general questions on how to handle Instants when working with them in a serialized object.

backend/src/main/java/de/aivot/gover/backend/utils/IsoTimestampUtils.java
```diff
@@ -3,55 +3,56 @@
...
-    public static String toUtcString(@Nonnull Instant value) {
-        return value.toString();
+    public static String nowWithOffset(@Nonnull ZoneId zoneId) {
+        return toOffsetString(Instant.now(), zoneId);
     }
 
     @Nonnull
-    public static String toUtcString(@Nonnull LocalDateTime value) {
-        return value.atOffset(ZoneOffset.UTC).toInstant().toString();
+    public static String toOffsetString(@Nonnull Instant value, @Nonnull ZoneId zoneId) {
```
mobaetzel, line 44: This method is often called with `ApplicationTimeZone.getZoneId()` as the zone id, maybe an overload without the zoneId could simplify some calls of this where when no zoneId is given, the default `ApplicationTimeZone.getZoneId()` is used.

backend/src/main/java/de/aivot/gover/backend/plugins/core/v1/javascript/DataObjectJavascriptV1.java
```diff
@@ -257,8 +259,8 @@ public ProxyObject update(@Nullable String schemaKey, @Nullable String itemId, @
     private static Map<String, Object> getItemData(@Nonnull DataObjectItemEntity entity) {
         var data = entity.getData();
         data.put("$id", entity.getId());
-        data.put("$created", entity.getCreated().toString());
-        data.put("$updated", entity.getUpdated().toString());
+        data.put("$created", IsoTimestampUtils.toOffsetString(entity.getCreated(), ApplicationTimeZone.getZoneId()));
+        data.put("$updated", IsoTimestampUtils.toOffsetString(entity.getUpdated(), ApplicationTimeZone.getZoneId()));
```
mobaetzel: Instead of always using the stringification of the entity timestamps, we could extend the method `mapToProxyObject` of the `JavascriptEngine` to handle Instants correctly by converting them to offset strings.

app/package.json
```diff
@@ -43,6 +43,7 @@
     "leaflet": "^1.9.4",
     "libphonenumber-js": "^1.13.8",
     "monaco-editor": "^0.52.2",
+    "luxon": "^3.7.2",
```
mobaetzel, line 44: I reckon, that this is the replacement for date-fns. I also reckon that removing the existing usages of date-fns is a huge chunk of work and should be done separately. We should add a clean-up task to the project management to do this before releasing version 5.

mobaetzel (changes requested):

backend/src/main/java/de/aivot/gover/backend/plugins/core/v1/nodes/actions/ApprovalActionNodeV1.java
danielkratz: Yes, that is correct. In general, we should keep temporal values as Instant in Java object graphs and let the configured ObjectMapper handle their JSON representation at the serialization boundary. For untyped Map<String, Object> data, the value will be read back as a string unless a typed target is provided, but there is no need to convert it before serialization. I changed this accordingly as ManualActionNodeV1 and DataChangeActionNodeV1 already handle processedAt the same way.

backend/src/main/java/de/aivot/gover/backend/utils/IsoTimestampUtils.java
danielkratz: Good suggestion. I added a toOffsetString(Instant) overload that defaults to the configured ApplicationTimeZone and updated the corresponding call sites. The existing overload accepting a ZoneId remains available for contexts that intentionally use a different zone, such as the business-time calls in DateJavascriptV1.

backend/src/main/java/de/aivot/gover/backend/plugins/core/v1/javascript/DataObjectJavascriptV1.java
danielkratz: I moved the Instant conversion into JavascriptEngine.mapToProxyObject and collectionToProxyArray, so timestamps remain typed until they cross the JavaScript boundary. I removed the explicit formatting from DataObjectJavascriptV1 and the equivalent code in CodeListJavascriptV1.

app/package.json
danielkratz: You are right that the migration was initially sizable. I had put the library cleanup on my mental todo list when introducing Luxon, but forgot to complete it. On this branch, however, nearly everything had already been migrated: compared to 38 source files on milestone/5.0.0, only six isolated usages remained. I migrated those remaining usages, added targeted tests for relative formatting and DST-sensitive dismissal behavior, and removed both date-fns dependencies.

mobaetzel (approved):

PR #156, opened 2026-09-01
feature: comunication provider

mobaetzel:
# Description
Adding communication providers to the system.


app/public/apple-touch-icon.png
danielkratz: Why were these files removed? These are the favicon and app icon for Prosuna and the PWA, which are still needed. This goes beyond the scope of a department theme.

app/public/favicon.ico
danielkratz: See comment on apple-touch-icon.

app/src/components/meta-element/meta-element.tsx
```diff
@@ -27,20 +30,6 @@ export function MetaElement({faviconUrl, title, titlePrefix}: MetaElementProps)
                 content="Prosuna – Die quelloffene Plattform für Ende-zu-Ende digitalisierte Verwaltungsprozesse. (prosuna.de)"
             />
 
-            <link
-                key="apple-touch-icon"
-                rel="apple-touch-icon"
-                sizes="180x180"
-                href={`${import.meta.env.BASE_URL}/apple-touch-icon.png`}
```
danielkratz: I definitely understand the reasoning here, but I think the best approach would be to separate the staff application from the customer pages. While the staff app does use the theme’s colors, the logo isn’t used at all, and the PWA requires an app icon. Prosuna, as an application, should continue to appear and be branded as Prosuna.

PR #157, opened 2026-09-08
feat(input-modes): Add dynamic configuration and typed authored values

danielkratz:
Integrate the dynamic input-mode prototype into process configuration and
establish an explicit authored/effective value boundary throughout the system.

## Changes

- Support Literal, Variable, No-Code and Low-Code modes for suitable
  backend-defined node configuration fields through explicit policies.
- Introduce typed authored-value wrappers while keeping effective values plain.
- Centralize mode resolution, dynamic-text interpolation and runtime validation,
  including nested replicating containers.
- Provide variable suggestions across all four sources and support custom paths.
  Suggestions explicitly do not guarantee runtime availability.
- Integrate dynamic text into Text and RichText fields, including syntax
  highlighting and cursor-based variable insertion.
- Reuse existing expression/script editors, preserve inactive mode drafts locally
  and add a side-by-side comparison in the Test Lab.
- Adapt forms, task views, uploads, PDFs, data objects, exports and configuration
  queries to the wrapper contract.
- Fix Monaco type hints, React Flow keyboard conflicts and No-Code editor layout.
- Clarify Low-Code data replacement, provide a data-preserving starter script
  and reduce the default inline editor height.
- Align icons, add project guidance and correct the security.txt expiration.
- Remove confetti animation from form submission. 

## Breaking Change

Authored values now use typed envelopes, including literal values and nested
authored rows. Raw authored values are no longer accepted by the API.

Apply migration `V30_0_0__authored_input_values.sql` and deploy the matching
frontend and backend together. The migration converts existing process-node,
payment-provider and storage-provider configurations.

Dynamic modes remain limited to explicitly enabled node configuration fields;
modeled forms cannot expose this capability.


backend/src/main/java/de/aivot/prosuna/backend/dataObject/services/DataObjectItemService.java
```diff
@@ -200,9 +200,7 @@ public DataObjectItemEntity performUpdate(@Nonnull DataObjectItemEntityId id,
     @Nonnull
     private Map<String, Object> deriveDataObjectItemData(@Nonnull DataObjectItemEntity entity,
                                                          @Nonnull DataObjectSchemaEntity schema) throws ResponseException {
-        var entityElementData = JsonMapperFactory
-                .getInstance()
-                .convertValue(entity.getData(), AuthoredElementValues.class);
+        var entityElementData = AuthoredElementValues.fromLiteralValues(entity.getData());
```
mobaetzel, line 203: The Method `AuthoredElementValues#fromLiteralValues(Map<String, ?>)` states, that for the replicating container layout another method should be used to transform the data. DataObjects may hold replicating container values; therefore, the conversion needs to be changed here.


Additionally, I've tested the data object item creation with a replicating container layout field and the first layer of the entity data is present as a Map of `String -> Object` but the child values of the replicating container items are in the form of `LiteralAuthoredInputValue`. The data arrives in the controller in this state where it is partially in the expected format of `Key -> Value` and partially in the target format of `AuthoredInputValue`. This contains a portion of ambiguity where a part of the data is in the correct format and is converted and another part is already in the right format. We should change this to either transform all data or transform none. For the sake of backend authority over formats, we should validate and transform the data on arrival and expect it to arrive as `String -> Object` because it makes working with this type of "Data Storing Objects" much easier. Additionally, we would solve all problems with Non-Literal AuthoredInputValues because we would need to verify that only literal authored input values are accepted by the endpoint and dismiss all other types of authored input values.


The conversion from `String -> Object` happens in the `DataObjectItemRequestDTO`.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/InputVariableResolver.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Nonnull
+    public Resolution resolve(@Nonnull InputVariableReference reference,
+                              @Nonnull ProcessExecutionData processExecutionData) {
+        validate(reference);
+
+        Object root = switch (reference.source()) {
+            case ProcessData -> processExecutionData.get(ProcessExecutionData.PROCESS_DATA_KEY);
+            case ProtectedProcessData -> processExecutionData.get(ProcessExecutionData.PROCESS_METADATA_KEY);
```
mobaetzel, line 47: Currently, the Identity-Objects in the Identities field of the protected process data are stored as `IdentityData`-Objects. This prevents the resolving of nested data when triggering the resolve from Java. Fix this by converting the children of the `IdentityDataMap` in the `ProcessDataService` during the `ProcessDataService#getInstanceData` from `IdentityData` to `Map<String, Object>`.

app/src/components/rich-text-input-component/rich-text-input-component.tsx
```diff
@@ -168,6 +238,38 @@ export function RichTextInputComponent(props: RichTextInputComponentProps) {
...
+            if (savedSelection == null) {
+                editorRef.current?.focus(() => editorRef.current?.insertMarkdown(markdown), {preventScroll: true});
+                return;
+            }
+
+            savedSelection.editor.setEditorState(savedSelection.editorState);
+            savedSelection.editor.focus(() => editorRef.current?.insertMarkdown(markdown));
+        },
+        insertVariableReference: (reference) => {
+            const placeholder = `{{ ${reference} }}`;
```
mobaetzel, line 258: We have a problem with the styling of our destination keys. They are in the format of `$.personen.0.name` but if we insert them into template tags (`{{ ... }}` or `{% ... %}`, they will be evaluated as JavaScript which is not valid JS. Either we change the syntax for destination keys or transform them before inserting them.

mobaetzel (changes requested):

backend/src/main/java/de/aivot/prosuna/backend/elements/models/elements/BaseInputElement.java
```diff
@@ -26,6 +28,12 @@ public abstract class BaseInputElement<T> extends BaseFormElement implements Inp
     @Nullable
     private String destinationKey;
 
+    @Nullable
+    private InputModePolicy inputModePolicy;
+
+    @Nullable
+    private DynamicTextPolicy dynamicTextPolicy;
```
mobaetzel, line 35: Why is the dynamicTextPolicy an attribute for all input elements when only the text input element and the rich text input element make use of it (as per ElementPOJOMapper Line 417)? Maybe I've missed some other places where this is used.

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
```diff
@@ -8,14 +11,78 @@
...
  */
-public class AuthoredElementValues extends HashMap<String, Object> implements Cloneable {
+public class AuthoredElementValues extends HashMap<String, AuthoredInputValue> implements Cloneable {
+    public static final String LITERAL_VALUE_PROPERTY = "value";
+
+    /**
+     * Wraps one map level. Replicating-container rows require the element-tree-aware conversion in
+     * {@code AuthoredInputValueService} when their nested values are effective values.
+     */
+    public static AuthoredElementValues fromLiteralValues(Map<String, ?> values) {
```
mobaetzel, line 28: Why does this method exist, if it is incompatible with all element values based on an element structure with replicating container layouts?

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
```diff
@@ -8,14 +11,78 @@
...
+
+    public Object getLiteral(String key) {
+        var value = get(key);
+        return value instanceof LiteralAuthoredInputValue literal ? literal.value() : null;
+    }
+
+    /**
+     * Unwraps one map level and rejects dynamic values. Nested authored rows intentionally remain structured.
+     */
+    public Map<String, Object> toLiteralValues() {
```
mobaetzel, line 75: Why does this exist if it is not recursive and does not unwrap the child values of replicating container layouts?

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
```diff
@@ -25,7 +92,15 @@ public AuthoredElementValues clone() {
         return clone;
     }
 
-    private static Object cloneValue(Object value) {
+    private static AuthoredInputValue cloneValue(AuthoredInputValue value) {
```
mobaetzel, line 95: Why does this method only clone literal values? Other types will not produce valid clones. Additionally, there is now `MapUtils#deepCopyValue` which should be used, if it is equivalent.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
```diff
@@ -0,0 +1,171 @@
...
+        var value = getLiteral(values, key);
+        return value == null ? null : jsonMapper.convertValue(value, type);
+    }
+
+    public void putLiteral(@Nonnull AuthoredElementValues values, @Nonnull String key, @Nullable Object value) {
+        values.put(key, literal(value));
+    }
+
+    @Nonnull
+    public AuthoredElementValues toLiteralAuthoredElementValues(@Nullable Object value) {
```
mobaetzel, line 59: This method does not resolve recursive authored element values correctly because it only puts the literals in the result. It is only called inside this service and should be private to prevent it from being unknowingly called in the wrong context.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
```diff
@@ -0,0 +1,171 @@
...
+@Service
+public class AuthoredInputValueService {
+    private final JsonMapper jsonMapper;
+
+    public AuthoredInputValueService(JsonMapper jsonMapper) {
+        this.jsonMapper = jsonMapper;
+    }
+
+    @Nonnull
+    public LiteralAuthoredInputValue literal(@Nullable Object value) {
```
mobaetzel, line 34: Why is this method public? it seems just like an internal helper.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
```diff
@@ -0,0 +1,171 @@
...
+        return new LiteralAuthoredInputValue(value);
+    }
+
+    @Nullable
+    public Object getLiteral(@Nonnull AuthoredElementValues values, @Nonnull String key) {
```
mobaetzel, line 39: Why is this method public? it seems just like an internal helper.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
```diff
@@ -0,0 +1,171 @@
...
+        return value instanceof LiteralAuthoredInputValue literal ? literal.value() : null;
+    }
+
+    @Nullable
+    public <T> T getLiteral(@Nonnull AuthoredElementValues values, @Nonnull String key, @Nonnull Class<T> type) {
+        var value = getLiteral(values, key);
+        return value == null ? null : jsonMapper.convertValue(value, type);
+    }
+
+    public void putLiteral(@Nonnull AuthoredElementValues values, @Nonnull String key, @Nullable Object value) {
```
mobaetzel, line 54: Why is this method public? it seems just like an internal helper.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
```diff
@@ -0,0 +1,171 @@
...
+        return unwrapLiteral(values.get(key));
+    }
+
+    @Nullable
+    public Object unwrapLiteral(@Nullable AuthoredInputValue value) {
+        return value instanceof LiteralAuthoredInputValue literal ? literal.value() : null;
+    }
+
+    @Nullable
+    public <T> T getLiteral(@Nonnull AuthoredElementValues values, @Nonnull String key, @Nonnull Class<T> type) {
```
mobaetzel, line 49: This method seems unused inside the project. Maybe a case of YAGNI :)

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return authoredRows;
+    }
+
+    @Nonnull
+    public Map<?, ?> toMap(@Nonnull Object value) {
+        return value instanceof Map<?, ?> map ? map : jsonMapper.convertValue(value, Map.class);
+    }
+
+    public void mapLiteral(
```
mobaetzel, line 161: The sole purpose of this method is being a helper for a single test case. this might not be ideal and should be removed.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
```diff
@@ -0,0 +1,171 @@
...
+
+            authoredRows.add(new ReplicatingContainerLayoutElementValue()
+                    .setId(row.get("id") instanceof String id ? id : null)
+                    .setValues(rowValues));
+        }
+        return authoredRows;
+    }
+
+    @Nonnull
+    public Map<?, ?> toMap(@Nonnull Object value) {
```
mobaetzel, line 157: Why is this method public? it seems just like an internal helper.

backend/src/main/java/de/aivot/prosuna/backend/process/models/ProcessNodeDefinition.java
```diff
@@ -161,21 +172,44 @@ default AuthoredElementValues prefillConfigurationOnImport(@Nonnull AuthoredElem
     }
 
     /**
-     * Validates the configuration of a process definition node entity. This used to check if the configuration of a node is valid before saving it or publishing the parent
-     * process. When errors are returned, the flag {@link ProcessNodeEntity#setSavedWithErrors(Boolean)} is set. Nodes can be saved with errors but a process cannot be published
-     * when at least one node does not validate correctly.
+     * Legacy authoring validation hook. It remains the default for providers whose configuration contains only
```
mobaetzel, line 175: if this is legacy, we should remove it.

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
```diff
@@ -8,14 +11,78 @@
...
+                Objects.requireNonNull(value, "An authored element value must use an input-mode envelope.")
+        );
+    }
+
+    @Override
+    public void putAll(Map<? extends String, ? extends AuthoredInputValue> values) {
+        values.forEach(this::put);
+    }
+
+    public Object getLiteral(String key) {
```
mobaetzel, line 67: Maybe we can modify this to return an Optional to make empty checks easier?

mobaetzel (changes requested):

app/src/models/input-mode.ts
```diff
@@ -0,0 +1,66 @@
+import {type NoCodeOperand} from './functions/no-code-expression';
+import {type ProcessNodeEntity} from '../modules/process/entities/process-node-entity';
+
+export type InputMode = 'Literal' | 'Variable' | 'NoCode' | 'LowCode';
+export type InputVariableSource = 'ProcessData' | 'ElementData' | 'ElementMetadata' | 'ProtectedProcessData';
```
mobaetzel, line 5: Why aren't these enums?

app/src/components/input-mode-field/input-mode-field.tsx
```diff
@@ -0,0 +1,785 @@
...
+    fieldProps: InputModeLiteralFieldProps;
+}
+
+export interface InputModeFieldProps<T> extends Omit<FormFieldLayoutProps, 'labelAction'> {
+    label: string;
+    hint?: string;
+    error?: string;
+    required?: boolean;
+    disabled?: boolean;
+    readOnly?: boolean;
```
mobaetzel, line 84: If we now support `readOnly` for input fields, maybe we can extend the `BaseInputElement` model with the new `readOnly` property to allow for a wider variety of configs?

app/src/components/code-editor/configure-monaco.ts
```diff
@@ -0,0 +1,28 @@
...
+
+// Use Monaco from the application bundle and let Vite emit each language service as a worker.
+self.MonacoEnvironment = {
+    getWorker(_, label) {
+        if (label === 'json') {
+            return new jsonWorker();
+        }
+        if (label === 'css' || label === 'scss' || label === 'less') {
+            return new cssWorker();
+        }
```
mobaetzel, line 17: This is not a real use case in Prosuna. We should remove this worker.

app/src/pages/staff-pages/settings/components/application-settings/application-settings.tsx
```diff
@@ -705,11 +707,11 @@ export function ApplicationSettings() {
                         currentGroup != null &&
                         <ElementDerivationContext
                             element={currentGroup}
-                            authoredElementValues={config}
+                            authoredElementValues={authoredConfig}
                             disabled={inputsDisabled}
                             onAuthoredElementValuesChange={(updated) => {
                                 console.log(updated);
```
mobaetzel, line 713: Unnecessary `console.log` here

app/src/modules/elements/components/element-derivation-context.tsx
```diff
@@ -469,13 +477,14 @@ function patchComputedElementStatesWithAuthoredValues(
...
+        const optimisticEffectiveValue = authoredValue?.type === 'Literal' ? authoredValue.value : null;
+        effectiveValues[currentElement.id] = optimisticEffectiveValue;
 
         if (isReplicatingContainerLayout(currentElement)) {
             nextElementStates[currentElement.id] = {
                 ...nextElementStates[currentElement.id],
-                subStates: Array.isArray(authoredValue) ?
-                    authoredValue.map((row, index) => {
+                subStates: Array.isArray(optimisticEffectiveValue) ?
+                    optimisticEffectiveValue.map((row, index) => {
```
mobaetzel, line 487: This replicating container layout value processing expects the items of the replicating container layout to be unpacked and not `AuthoredInputValue`.

app/src/modules/elements/components/element-derivation-context.tsx
```diff
@@ -469,13 +477,14 @@ function patchComputedElementStatesWithAuthoredValues(
                 valueSource: ComputedElementValueSource.Authored,
             },
         };
-        effectiveValues[currentElement.id] = authoredValue;
+        const optimisticEffectiveValue = authoredValue?.type === 'Literal' ? authoredValue.value : null;
+        effectiveValues[currentElement.id] = optimisticEffectiveValue;
```
mobaetzel, line 481: This seems as computed effective values (produced by `authoredValue.type != 'Literal'` eg a no code computed effective value) to always be dropped for null. Is this correct?

mobaetzel (dismissed):

backend/src/main/java/de/aivot/prosuna/backend/dataObject/services/DataObjectItemService.java
danielkratz: You're right: I was only unwrapping and rewrapping the outer level, leaving the repeated rows in a different format. I've followed your suggestion and changed the request to plain `String -> Object` data. The backend now uses the schema-aware conversion for all levels before derivation and validation.

I've also changed the response to plain values so that reading and writing data objects use the same contract. Authored wrappers now stay inside the editor draft and the backend derivation. Incoming objects are treated as data, never as executable input-mode instructions.

I've added tests for nested containers, row IDs, create/update round trips, validation failures, and editor save/reset behavior. The focused tests, typecheck, and both frontend builds pass.

Commit: `e4f775dec`

backend/src/main/java/de/aivot/prosuna/backend/elements/services/InputVariableResolver.java
danielkratz: I've followed your suggestion and converted the identity records to plain maps in `ProcessDataService#getInstanceData`, using the injected JSON mapper. The shared execution snapshot now lets the Java resolver traverse nested identity attributes, while the persisted `IdentityDataMap` stays unchanged.

I've added a regression test for nested attributes, UUID normalization, and the distinction between missing values and explicit nulls. All 14 focused backend tests pass.

Commit: `03cc8afad`

app/src/components/rich-text-input-component/rich-text-input-component.tsx
danielkratz: I’d like to clarify which direction you’d prefer here. Since we also use destination keys in variable bindings, No-Code, write targets, and form mappings, I see two options:

- Keep `personen.0.name` as the destination-key notation and convert it to `$.personen[0].name` when inserting a reference into JavaScript, including template expressions such as `{{ $.personen[0].name }}`.
- Use `personen[0].name` consistently for concrete paths in configuration fields as well. The same path could then be used as `$.personen[0].name` in Low-Code and `{{ $.personen[0].name }}` in templates. This would mean updating the path readers, writers, validation, and existing configuration paths together.

I’m leaning toward the second option so users don’t have to switch between `.0.` and `[0]` depending on where they use a reference. Wildcard paths would still need an explicit distinction: for example, `personen[*].name` could represent a path pattern in mappings, but it would not be a valid JavaScript property access.

Would you prefer that broader alignment, or would you keep the current destination-key notation and convert references at the JavaScript insertion boundary?

backend/src/main/java/de/aivot/prosuna/backend/elements/models/elements/BaseInputElement.java
danielkratz: You haven’t missed another use case: dynamic text is only intended for Text and RichTextInput. I originally put the policy on BaseInputElement to simplify the shared mapper and derivation code, but I agree that this exposed it too broadly.

I’ve moved the policy to the two supported element types and introduced a small `DynamicTextElement` interface for shared handling. I’ve also aligned the frontend models. Overrides still use the policy from the original definition and cannot enable interpolation themselves. The JSON structure of supported fields remains unchanged.

I’ve added regression tests for serialization, equality, unsupported inputs, overrides, and frontend rendering.

Commit: `5b04d5c76`

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
danielkratz: I agree that this was too easy to use as a general conversion method. Its remaining production caller was the Approval draft fallback, whose modeled task view can also contain repeated rows.

I’ve switched that fallback to the existing schema-aware conversion and removed `fromLiteralValues`. The fallback remains supported; the regression test covers nested rows, row IDs, explicit nulls, and ordinary business objects that must not receive additional envelopes. All 19 focused tests pass.

Commit: `4d561e871`

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
danielkratz: You’re right: unwrapping only the outer level does not produce a complete effective-value structure. This method was only used by tests, so I’ve removed it along with the authored-to-runtime test shortcut.

The affected tests now construct the expected effective fields explicitly. The controller tests receive effective results separately from normalized authored inputs, and the mapping tests build distinct authored and effective container rows with matching IDs. This avoids keeping an incomplete conversion API in production code. All 152 selected backend tests pass.

Commit: `3e20d5073`

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
danielkratz: I agree that the clone was incomplete: No-Code operands and typed replicating-container rows still shared mutable state with the original. I've fixed that in `2e6a5f1ff` and added tests that modify the copy and check that the original stays unchanged.

I've also reused `MapUtils.deepCopyValue` through a small type-specific copy hook. It wasn't equivalent on its own: it left operands and typed rows untouched and turned nested `AuthoredElementValues` into ordinary maps. The hook keeps the domain handling outside `MapUtils` while sharing its collection and array traversal.

Variable and Low-Code records remain shared intentionally because their contents are immutable. For No-Code, the copy preserves raw draft fields rather than normalizing them through getters or JSON. The focused backend checks passed (122 tests).

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
danielkratz: I've revisited the fallback I mentioned above and removed it in `7e5daaac6`. The current Approval flow already uses the shared autosave contract under `staffTaskViewData`; the fallback only supported the old top-level runtime-data format, which we don't need to retain.

Approval now uses the inherited reader directly. I've replaced the legacy-format test with a regression test showing that unrelated runtime data does not override the initial form values. The existing canonical saved-draft test remains, and all 19 focused tests pass.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
danielkratz: I agree: this only wraps one level and shouldn't be callable as a complete conversion. I've made it private and renamed it `wrapLiteralLevel`, with a comment explaining that container children must be normalized using the element tree. The test now uses the public schema-aware conversion instead. Commit: `3701a0826`.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
danielkratz: You're right, `toMap` is only an internal helper. I've made it private; none of the production callers need direct access to it. The schema-aware conversion remains public, and all 91 focused tests pass. Commit: `3701a0826`.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
danielkratz: You're right that this doesn't need a public service method. I've removed `literal` altogether: its only production use was through another redundant helper, and wrapping now uses the existing `AuthoredElementValues.putLiteral` method. Commit: `aa1d8e06e`.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
danielkratz: I've checked the callers and agree that this service-level `getLiteral` isn't needed. I've removed it rather than retaining another private wrapper. The value model already supports key-based literal access; `unwrapLiteral` remains because derivation uses it to inspect individual envelopes. Commit: `aa1d8e06e`.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
danielkratz: I agree. I've removed the service helper and changed the internal conversion to use `AuthoredElementValues.putLiteral` directly. This avoids maintaining two methods for the same operation. Commit: `aa1d8e06e`.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
danielkratz: Agreed, this was speculative convenience with no caller. I've removed the typed `getLiteral` overload without adding a replacement. Commit: `aa1d8e06e`.

backend/src/main/java/de/aivot/prosuna/backend/elements/services/AuthoredInputValueService.java
danielkratz: You're right: keeping `mapLiteral` just for that test wasn't justified. I've removed it and replaced the helper-oriented test with coverage of the supported `unwrapLiteral` operation, including null and dynamic values. All 100 tests in the final focused run pass. Commit: `aa1d8e06e`.

backend/src/main/java/de/aivot/prosuna/backend/process/models/ProcessNodeDefinition.java
danielkratz: I agree that we don't need to retain the legacy hook. I've migrated the remaining implementations and tests to the context-based method and removed the old overloads in `5f5b76f3b`.

Form and webhook validation remain explicitly authoring-only, while the existing runtime checks for dynamic AI and storage configurations are preserved. I've also removed forwarding helpers and empty overrides, retained the relevant TODOs, and documented the phase responsibilities. All 99 selected backend tests pass, including checks that runtime form and webhook validation doesn't query definition uniqueness.

backend/src/main/java/de/aivot/prosuna/backend/elements/models/AuthoredElementValues.java
danielkratz: I've checked the callers, and I'd prefer to keep the nullable payload here. Most pass it to nullable formatters or inspect its runtime type, so an `Optional<Object>` would mostly be immediately unwrapped again.

It would also keep the existing ambiguity between a missing entry, a null literal, and a non-literal mode; callers needing that distinction must inspect the envelope. I don't see enough benefit in the current usages to justify changing the return type, so I've left `getLiteral` unchanged.

app/src/components/input-mode-field/input-mode-field.tsx
danielkratz: I agree that read-only elements would be useful. The current `readOnly` property controls the input-mode UI, but it is not yet a complete element-level contract.

I'd take this up in our existing element harmonization task (Element-Harmonisierung), so I can align the backend/frontend models, editor and views together. In particular, I'd like to define how read-only differs from disabled for focus, inspecting/copying values and value derivation, where disabled already affects whether authored or calculated values take precedence. That seems a better fit than exposing a flag here that only some controls fully support.

app/src/components/code-editor/configure-monaco.ts
danielkratz: I see your point about not shipping a worker without a real use case. CSS is currently offered as a language in the CodeInput element editor, though, and `CodeInputView` passes that selection through to Monaco. Removing the worker would therefore reduce the editor support for an option that is already available.

I'd also been thinking ahead to a template editor that could edit or inject CSS. That part may well be my newly learned YAGNI lesson arriving right on cue.

Given the existing CSS language option and that planned direction, would you still like me to remove the CSS worker? If so, should I also remove CSS from the language options, or deliberately keep it with reduced editor support?

app/src/models/input-mode.ts
danielkratz: You're right, there wasn't a reason to keep these as string unions. I've changed both `InputMode` and `InputVariableSource` to string enums and updated their usages in the wrappers, policies, controls and tests in 6d18f5800. The JSON identifiers stay unchanged; I've added contract tests for that as well.

app/src/pages/staff-pages/settings/components/application-settings/application-settings.tsx
danielkratz: I've removed this log and the adjacent tab-change log in c8062bb9c. Neither was needed for the settings workflow.

app/src/modules/elements/components/element-derivation-context.tsx
danielkratz: I've corrected the nested conversion in c73a1aba0. The outer Literal was already unwrapped, but each row's `values` could still contain authored envelopes, which were copied into the optimistic effective data.

I now project the child values recursively using the element schema. Previous effective values and states are matched by row ID, so reordering or deleting rows doesn't move derived or identity values to another row. Ordinary object payloads remain untouched. I've added coverage for nested containers, row changes, protected child values and input immutability.

The separate question about clearing dynamic results is still open; I haven't changed that policy in this fix.

app/src/modules/elements/components/element-derivation-context.tsx
danielkratz: You're right: this cleared the effective result of every editable dynamic field, even when only an unrelated field changed. Another derivation was not guaranteed either, since the change handler only looked at referencedIds.

I've corrected this in 9866b237f. I now keep existing results and states for unchanged definitions, invalidate an edited expression, and trigger backend derivation even without referencedIds. Nested definitions are compared by row ID, so reordering doesn't invalidate the wrong result. Switching back to Literal also triggers derivation.

The backend still deliberately defers runtime evaluation in authoring contexts, so a null result there remains valid. I've covered all three dynamic modes, unchanged dynamic containers, reordered rows, mode switches, and null/false/zero responses with tests.

app/src/components/rich-text-input-component/rich-text-input-component.tsx
mobaetzel: I'm completely on the same page. Standardizing on $.personen[0].name for both destination key bindings and low-code references will significantly reduce cognitive load for users. While this requires refactoring our existing services to handle reading, writing, and payload building with these keys, it's a necessary trade-off for the improved user experience.

app/src/components/code-editor/configure-monaco.ts
mobaetzel: You've put a lot of care into the app's styling and colors. If we're definitely planning to let users customize these in the future, we should retain this configuration. I agree with keeping it.

app/src/components/rich-text-input-component/rich-text-input-component.tsx
danielkratz: I have implemented the broader alignment we agreed on. A destination key such as `personen[0].name` now uses the same notation as `$.personen[0].name` in Low-Code and `{{ $.personen[0].name }}` in templates, without an insertion-time conversion.

I updated the shared readers/writers, payload transformation, variable and No-Code validation, mapping and data-type-validation nodes, replication metadata, dependency lookup, and frontend path editors. `personen[*].name` remains a mapping pattern with contextual indices, not an executable JavaScript reference. The old dotted numeric/wildcard notation is rejected rather than retained as a fallback.

I also checked the SQL examples and migrations and found no old array paths there to rewrite. The changed test fixtures now use bracket notation, with explicit rejection tests for the old syntax and a write-to-template integration test.

Committed as `4aeb07e30` and `73eb0bd77`. Verification: 1,238 backend tests, 754 frontend tests, typecheck, and both staff/customer builds passed.

jp-schroeder (approved):

